PR #28, opened 2018-09-06
add pca10059

dhalbert:
Add support for pca10059 dongle.

Also some whitespace fixes (happened by accident when emacs cleaned up trailing spaces).


dhalbert:
To load, I used a J-Link. It appears you need to load the softdevice:
```
make BOARD=pca10059 erase
make BOARD=pca10059 sd
make BOARD=pca10059 flash
```

I forgot `sd` at first and it blinked but did not show NRF52BOOT.

hathach:
that is a very helpful feature of emacs.

arturo182:
@hathach what editor do you use?

hathach:
I used eclipse + emac-plug plugin for writing most of the code. Emacs is great but I found eclipse perspective is better for my taste.

hathach:
There is lots of great editor, but I am too lazy to change :(

hathach:
> I forgot sd at first and it blinked but did not show NRF52BOOT.

Yeah, the **flash** only flash the bootloader, **sd** must be flashed first

arturo182:
@hathach Then you can configure Eclipse to strip whitespace too https://stackoverflow.com/a/2618521

hathach:
@arturo182 thanks, there is a lots of settings to do all sorts of stuffs with eclipse. It currently does that with .c and .h. Makefile is probably need some diving config.

PR #34, opened 2018-12-05
Rework leds, add Travis and add Particle boards

tannewt:
The LEDs now match the atmel uf2 bootloader by ramping the single LED up and down at different speeds. The RGB pixel is used for color status when available as well.

Travis will now automatically build all boards and upload artifacts to a release. (Release untested at the moment.)

There is also a fix for the NeoPixel code to correct writing to the last two spots in the output. It was writing past the end of the array.


hathach:
uhmmm, I got an error when trying to build, maybe some makefile variable is not update

```
$ make BOARD=pca10056 clean all
make: *** No rule to make target '_build-pca10056/pca10056_bootloader-99fb105.out', needed by 'all'.  Stop.
```

tannewt:
@dhalbert Updated but left in to match the style of the repo.

@hathach Yup! Missed one spot. Updated now.

hathach:
@tannewt thanks, I could build now, but the flash command doesn't work, it seems to miss the new output name with "-nosd" suffx

```
hathach@ubt18:~/Documents/Adafruit_nRF52_Bootloader$ make BOARD=pca10056 all flash

   text	   data	    bss	    dec	    hex	filename
  28104	    344	  19774	  48222	   bc5e	_build-pca10056/pca10056_bootloader-73a9d50-nosd.out

make: *** No rule to make target '_build-pca10056/pca10056_bootloader-73a9d50.hex', needed by 'flash'.  Stop.
```

tannewt:
@hathach Ok I fixed the flash and genhex targets as well.

hathach:
@tannewt Superb! I tested, and it work greats. The new blinky is less aggressive. One final changes, can you remove the r0 trailing in the binary name **feather_nrf52832_bootloader-689c577_s132_6.1.1`r0`.zip** . It is intended to minor increase revision update (bug fix) while having the same SD version. Since you include the git hash, there is no real reason to keep it anymore :D. 

PS: The pattern for Blue LED (secondary for nrf52832) in OTA mode when it is connected vs not-connected is a little bit hard to distinguish if user don't pay attention to it. Otherwise everything is perfect.

tannewt:
Ok, I've removed the revision. I'm going to merge and do a release.

PR #36, opened 2018-12-19
fixed the incosnsitent led pattern by using systick to call led_tick()

hathach:
@tannewt should be the reviewer but is currently on vacation. I will merge this right away since we need to release bsp update for Arduino asap


ladyada:
hiya i took a look at this
if tannewt's not around, you can always have dan or i review - even if we're not experts at the code, its good to have another person look :)

tannewt:
I'm not sure we want this because it will hide any hangs. Having the tick in the main loop means it will only update if things are working well. With SysTick it will continue even if we hard fault into an infinite loop because interrupts may still work during an infinite loop.

Flash timing consistency isn't as important to me as demonstrating code is running as expected. That is why I moved the blinking off the pwm peripheral as well.

hathach:
The led pattern is really important, since it is what Arduino user will look at every upload. Previous one is too random. The led tick doesn't help much when hitting infinite loop either (for hardfault, it will reset), if you like a fail-safe, we can add watchdog for that.

tannewt:
Ok, this is fine then.

PR #40, opened 2019-01-10
Fix OTA using nRF toolbox on iPhone (tested on SE and XS)

se7ensong:
When using nRF Toolbox on iPhone XS for legacy DFU, it sends PHY UPDATE REQ but are not handled properly causing update to stuck.


hathach:
@se7ensong thank you, I have just confirmed your PR with an "borrowed" iPhone X.

hathach:
hmm, just realize you changed HCI_RX_BUF_QUEUE_SIZE from 4 to 8, it is not related to the issue. I will change it back

PR #44, opened 2019-01-31
merge from Develop

### LINKED ISSUE #41: Hardcoded RXD/TXD pin
For the serial bootloader mode the RXD and TXD pins are hardcoded in sdk_config.h and changes in the boards/ folder do not propagate to the nordic SDK part of the bootloader.
I propose using the RXD/TXD/RTS/CTS pin definitions from the board headers in the sdk_config.h

hathach:
- Fixed PWM psel[1] is not reset
- Fixed #41 move RXD, TXD into board header
- Added Metro nRF52840 Rev A
- Fixed #40 OTA issue with BLE_GAP_EVT_PHY_UPDATE_REQUEST e.g connecting with iPhone X


hathach:
thanks @dhalbert for testing this out. OTA should work as before, this fixes an issue with latest central such as iPhone X which is capable of 2M Phy and requesting nrf52840 to negotiate.

PR #45, opened 2019-02-15
increase buffer depth to fix OTA issue with nrfConnect on iOS

hathach:
OTA issue discussed in forum
https://forums.adafruit.com/viewtopic.php?f=57&t=123583&p=729259#p729259

Reason is latest iOS and updated SD support multiple packets per connection interval (max 6 packets), which is quickly overflow the current buffer depth (previously = 4). Buffer depth is increased to 8 now. For Android, it doesn't seem to support multiple packets yet (at least on my 8.0.0 )

- also add optional debug helper
- increase led conn cycle when ble connected for easier recognition

### DIFF
--- a/lib/sdk11/components/libraries/bootloader_dfu/dfu_transport_ble.c
+++ b/lib/sdk11/components/libraries/bootloader_dfu/dfu_transport_ble.c
@@ -736,6 +736,11 @@ static void on_ble_evt(ble_evt_t * p_ble_evt)
     uint32_t                              err_code;
     ble_gatts_rw_authorize_reply_params_t auth_reply;
 
+#ifdef CFG_DEBUG
+    extern const char* dbg_ble_event_str(uint16_t evt_id);
+    ADALOG("BLE", dbg_ble_event_str(p_ble_evt->header.evt_id));
+#endif
+
     switch (p_ble_evt->header.evt_id)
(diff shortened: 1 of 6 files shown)


lib/sdk11/components/libraries/bootloader_dfu/dfu_transport_ble.c
```diff
@@ -1088,3 +1094,109 @@ uint32_t dfu_transport_ble_close()
...
+  return NULL;
+}
+
+/*------------------------------------------------------------------*/
+/* BLE Event String
+ *------------------------------------------------------------------*/
+static lookup_entry_t const _strevt_lookup[] =
+{
+    // BLE common: 0x01
+    { .key = BLE_EVT_USER_MEM_REQUEST                , .data= "BLE_EVT_USER_MEM_REQUEST"                },
```
dhalbert: You could probably do this with a preprocessor macro, so you avoid typing the name twice (and possibly introducing a typo). But not a big deal.

hathach: ah yeah, you are right, I will do that later on :)

hathach:
Thanks @dhalbert for approving, I will merge it now :)

PR #49, opened 2019-03-04
Support the MDK nrf52840 USB Dongle.

gpshead:
This dongle doesn't have usable buttons, so this UF2 bootloader is primarily usable via the double reset button press.

I didn't want to consume two of the few pins on the dongle for fake button jumpering as I assume people want to use those in their application.


hathach:
Thanks for the PR, though without both of buttons, it won't be able to force bootloader into OTA mode. It is still possible to trigger OTA while it is in application mode though.

PR #53, opened 2019-03-12
Fix signed/unsigned bug.

henrygab:
Without this fix, any error will fail to be detected.
Specifically, because both ```resplen``` and ```bufsize``` were both ```uint16_t```, they could never store a negative value.  ```resplen``` was set to a negative value (```0xFFFF```), to indicate an error at line 93.

Then, line 101:
```if ( resplen > bufsize ) resplen = bufsize;```
would change the ```resplen`` from the error value of ```0xFFFF``` to the non-error value ```bufsize```, and TinyUSB would treat this as success.

The fix changes ```resplen``` from ```uint16_t``` to ```int16_t```, allowing it to store negative values, and explicitly upcasts ```bufsize``` from unsigned 16-bit to signed 32-bit value for the error check.


hathach:
Superb !! You have an eagle eye there :+1:

PR #54, opened 2019-03-12
Avoid returning uninitialized memory

henrygab:
There are many, many code paths where SCSI commands such as READ will return uninitialized memory to the host.

Rather than fixing the many locations this can occur, fix the dispatch to zero the memory buffer at the start of the command.


hathach:
super safe coding, indeed

PR #55, opened 2019-03-12
Use local variable to store partially-calculated size.

henrygab:
This avoid re-entrancy problems, such as if this is ever
interrupted mid-execution, and called a second time before
the initial execution completes.


hathach:
thank you very much for the thoughtful PR, I have no idea this could happen :D

PR #56, opened 2019-03-13
Ignore additional SES and J-Link files

henrygab:


.gitignore
```diff
@@ -54,3 +54,5 @@ dkms.conf
 src/segger/Output
 _build-*/
 bin/
+*.emSession
+*.jlink
```
hathach: there is a warning for no newline at end of file. Can you just insert one, I am not sure how important it is, but just better to be safe since people use different OS could have issue with this.

henrygab: OK, done.  How important?

The newline is viewed by some software as a line terminator.  Some software processes files line-by-line, and consider the lack of a line terminator to mean that the final bytes are not a fully defined "line".  They fail to check for the count of bytes read at the end of a stream, and thus doesn't process the final line.

_(Personally, I've always considered it a bug to not process the trailing bytes of a file as a line.)_

Let me know if anything else should be updated.

hathach:
That is good enough, thanks for the PR :)

PR #64, opened 2019-04-12
Add support for Electronut Labs Papyr

ntavish:
Circuitpython port here: https://github.com/adafruit/circuitpython/pull/1798

### DIFF
--- a/src/usb/usb_desc.c
+++ b/src/usb/usb_desc.c
@@ -79,30 +79,33 @@ enum {
 // STRING DESCRIPTORS
 //--------------------------------------------------------------------+
 
+#ifndef USB_STRING_DESCRIPTORS
+#define USB_STRING_DESCRIPTORS {                                                                                 \
+    /* 0: is supported language = English */                                                                        \
+    TUD_DESC_STRCONV(0x0409),                                                                                       \
+                                                                                                                    \
+    /* 1: Manufacturer */                                                                                           \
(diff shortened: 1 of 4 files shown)


ntavish:
Papyr's current draw sometimes spikes over 100mA, so there is a change of brownout/reset. Circuitpython was rebooting to safe mode if pins aren't set 'correctly' (seen this once or twice). That's why I included  `boards.h` inside `usb_desc.h`: to increase requested power, and to override PID. Let me know if this is not the preferred approach. I can in a separate PR also cleanup other boards if information about VID/PID is available.

hathach:
@ntavish does your board draw more than 100 mA in bootloader mode or application mode ? Though I agree we can allow `board` to overwrite several value of USB.

src/usb/usb_desc.c
```diff
@@ -79,30 +84,33 @@ enum {
 // STRING DESCRIPTORS
 //--------------------------------------------------------------------+
 
+#ifndef DEVICE_STRING_DESCRIPTORS
```
hathach: should be name `USB_STRING_DESCRIPTORS` instead

ntavish:
Oh you're right, this needs to be application mode :man_facepalming:. My excuse is that I was doing it at 3AM. The other modifications to the USB descriptors, the VID:PID for example are still required though.

PR #66, opened 2019-04-24
Clean up board and makefile

hathach:
fix #65 by forcing NFC as GPIO

PR #67, opened 2019-04-24
Update macros.h

henrygab:
Per continued discussions for #57.


hathach:
thanks @henrygab , you don't have to do this PR, I already included it as part of PR #66 . Since we are in hurry to wrap this up for new board metro nrf52840 and worry you may be busy meanwhile  :)

PR #69, opened 2019-05-07
Use either -Werror OR -fstack-usage (if STACK_USAGE=1) never both

pyro9:
-fstack-usage gives warnings whenever inline ASM is used
  -Werror makes that an error and causes make to fail

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -229,12 +229,18 @@ INC_PATHS = $(addprefix -I,$(IPATH))
 #flags common to all targets
 CFLAGS += -mcpu=cortex-m4
 CFLAGS += -mthumb -mabi=aapcs --std=gnu99
-CFLAGS += -Wall -Werror -Os -g3
+CFLAGS += -Wall -Os -g3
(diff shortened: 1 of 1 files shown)


hathach:
which build/board that causes the issue. I haven't experienced this, or maybe it is the gcc version you are running.

pyro9:
It happens with BOARD=feather_nrf52840_express using gcc version 5.2.1 20151202 (release) [ARM/embedded-5-branch revision 231848] (GNU Tools for ARM Embedded Processors) 
(The one distributed with the Adafruit BSP)

hathach:
your gcc version is too old, please try to upgrade to at least version 7 and test again. The included in bsp is also old, we plan to upgrade it as well when thing is stable.
 
```
gcc (Ubuntu 7.4.0-1ubuntu1~18.04) 7.4.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

hathach:
Closed since not reproducible

Issue #70, opened 2019-05-20
[Bug] only partial fix for known SDK11 bootloader bug when updating soft device


fanoush:
There is known bug in bootloader code when updating SoftDevice which is larger than previous one https://devzone.nordicsemi.com/f/nordic-q-a/16774/updating-from-s132-v2-0-x-to-s132-v3-0-0-with-dual-bank-bootloader-from-sdk-v11-0-0-does-not-work

You actually do have fix for this here
https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/master/lib/sdk11/components/libraries/bootloader_dfu/dfu_single_bank.c#L686

however same block alignment fix should be also here
https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/master/lib/sdk11/components/libraries/bootloader_dfu/dfu_single_bank.c#L792
this code is called only when recovering from power failure or reset so the chance of triggering this is low

And same fix should be in dual bank code.

hathach:
thanks @fanoush for bringing this up, I remembered reading the same topic on devzone as well. But my memory fades and is currently busy with other issue. It would be great If you could do a PR for this, if you are busy, I will do it as well later.

fanoush:
So I created pull request #71   I hope it would be somehow referenced here when mentioning issue number in PR name.

BTW thanks for keeping legacy unsigned DFU update alive. I used this bootloader to port micropython to DS-D6 nrf52 based fitness tracker and the dual serial/BLE DFU is awesome.

I made some other changes like support for only 1 button but I guess you are not interested in that.

However I have one question - why pstorage is configured to use 4 blocks here https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/master/src/pstorage_platform.h#L72 and  there is this assert https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/master/src/main.c#L125 and setting here https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/master/Makefile#L247 which both expects 7 blocks starting on specific 0x6D000 location? Are those two related (pstorage vs DFU_APP_DATA_RESERVED)? Then why the difference? Is this  0x6D000 also hardcoded elsewhere in other projects?

I guess I can remove this area from my bootloader fork and it will continue to work?

hathach:
[...]

PR #71, opened 2019-05-24
fix for issue #70

fanoush:
Fix also dfu_sd_image_validate method in single and dual bank code as per https://devzone.nordicsemi.com/f/nordic-q-a/16774/updating-from-s132-v2-0-x-to-s132-v3-0-0-with-dual-bank-bootloader-from-sdk-v11-0-0-does-not-work/185106#185106

This dfu_sd_image_validate code path is called when recovering from power failure or reset and incomplete SoftDevice copy needs to be finished.


hathach:
thank you very much for this PR

Issue #76, opened 2019-09-08
Circuit Playground Bluefruit forcing OTA DFU mode


khanning:
After applying the latest precompiled bootloader + SD for the Circuit Playground Bluefruit (`circuitplayground_nrf52840_bootloader-0.2.12_s140_6.1.1`), the device forces itself into OTA DFU mode (purple neopixel ring).  
  
Holding down either button while resetting allows the device to enter CDC/UF2 DFU mode. However, on subsequent resets the device forces itself back into OTA DFU mode so there doesn't seem to be a way to run application firmware after a reset.

**Set up**
  - Circuit Playground Bluefruit
  - Mac OSX 10.14.6

**To Reproduce**
1. Place Circuit Playground Bluefruit in CDC/UF2 DFU bootloader mode 
2. Apply new bootloader + SD  
`adafruit-nrfutil --verbose dfu serial --package circuitplayground_nrf52840_bootloader-0.2.12_s140_6.1.1.zip -p /dev/tty.usbmodemXX -b 115200 --singlebank --touch 1200`
3. After receiving success message, the device enters OTA DFU mode (purple neopixel ring)  
```
Starting DFU upgrade of type 3, SoftDevice size: 151016, bootloader size: 29200, application size: 0
Sending DFU start packet
Sending DFU init packet
Sending firmware file
########################################
########################################
########################################
########################################
########################################
########################################
########################################
########################################
################################
Activating new firmware

DFU upgrade took 19.310901165008545s
Device programmed.
```
4. Enter CDC/UF2 DFU mode by holding either button 1 or 2 and resetting the device.
5. Apply application firmware using either CDC or UF2.
6. The application runs properly.
6. Reset the device. It enters OTA DFU mode.

hathach:
should be fixed by pr #77 . Thank you very much for reporting and fixing the issue.

PR #77, opened 2019-09-09
Add check for button direction

khanning:
boards.c incorrectly assumes that an active low state indicates that a button is pressed.
```
bool button_pressed(uint32_t pin)
{
  return (nrf_gpio_pin_read(pin) == 0) ? true : false;
}
```
This is not true with devices like the Circuit Playground Bluefruit where the buttons have an active high state and causes the issue described in #76.

This PR adds a new def `BUTTON_DIR` to check _nrf_gpio_pin_read_ against. `BUTTON_DIR` is inferred from `BUTTON_PULL` which is already defined in the custom board files.

### DIFF
--- a/src/boards.h
+++ b/src/boards.h
@@ -38,6 +38,13 @@
 #ifndef BUTTON_FRESET
 #define BUTTON_FRESET   BUTTON_2
 #endif
+#ifndef BUTTON_DIR
+#if BUTTON_PULL == NRF_GPIO_PIN_PULLDOWN
+  #define BUTTON_DIR   1
+#elif BUTTON_PULL == NRF_GPIO_PIN_PULLUP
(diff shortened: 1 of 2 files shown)


src/boards.h
```diff
@@ -38,6 +38,13 @@
 #ifndef BUTTON_FRESET
 #define BUTTON_FRESET   BUTTON_2
 #endif
+#ifndef BUTTON_DIR
```
hathach: Thanks, but let's rename the `BUTTON_DIR` into `BUTTON_STATE_ACTIVE` 0 for active low, and 1 for active high, which is a little clearer.

We also don't need to define BUTTON_PULL, just check the BUTTON_STATE_ACTIVE and set up the opposite pullup in `button_init()` above

Issue #72, opened 2019-06-25
Upgrade to support Softdevice ver 7.x


hathach:

dhalbert:
I'm waiting for the doc of the changes, and would probably wait for at least 7.0.1 so they can iron out bugs. If there are upward-compatibility issues we have to be careful.

hathach:
agree, we should wait for the next release with bug fix. I will make branch to start adopt the new flash layout and testing.

hathach:
@dhalbert SD v7 seems to be required to support the new nrf52833 and probably nrf5340 later on as well. 
https://www.nordicsemi.com/Software-and-tools/Software/S140/Download#infotabs
![image](https://user-images.githubusercontent.com/249515/69826593-6fbeb000-1246-11ea-810e-1fd0634cd033.png)

dhalbert:
Thanks. We can discuss about an upgrade strategy next week.

hathach:
@dhalbert Please take the break, It is holiday in USA. I already wrapped it up with #88 . We will upgrade the SD later when **it supports 5340 dual core chip**. Since we probably make a feather for 5340 than 52833 :) .

PS: Next week I will take a whole week off for a vacation. Though I am still reachable via email (hopefully)

dhalbert:
@hathach Do we need to upgrade everything to 7.x.x, or can 6.x.x and 7.x.x coexist? I was going to build this for pca10100 just so I could use it to upload the SD on a pca10100 board but realized it was not updated yet. Thanks.

hathach:
@dhalbert it can be co-existed, the bootloader can work with either bootloader since they are decoupled. However, since the flash layout changes due to SD_SIZE, the application must has correct linker. Actually, I would suggest to include the Softdevice as part of the application uf2, although the uf2 is larger, it doesn't actually take time to flash since most of the time the softdevice matches with the flash contents. To sum up

- You can flash bootloader with/wihtout SD v6 as current state of bootloader
-  In both ways, just include SD v7 (or any later SD version) as part of circuitpython to uf2 (use mergehex before uf2conv), it should then just work for all the board.

Note: This need latest bootloader (master tip) which isn't released yet since I think we need a bit more testing. Hope this help.

pabigot:
[...]

Issue #24, opened 2018-08-23
MSC/UF2 to upgrade bootloader + SD


hathach:
So far only DFU Serial could upgrade Bootloader + SD combo, uf2 should also do this as well. With a set of specific requirement e.g
- family option is different than one used by application firmware
- SD + bootloader must be upgraded at the same time to make sure that are in sync and work as expected
- SD data contains must be check to make sure it is S140
- uf2 of bootloader contain VID/PID which is used to prevent upgrading to incompatible board.

dhalbert:
On atmel-samd UF2, the UF2 bootloader cannot update itself directly. Instead, you load a update program into non-bootload (user) space which then rewrites the bootloader. See the update-bootloader*.ino and .uf2 files in the UF2 release: https://github.com/adafruit/uf2-samdx1/releases/latest

hathach:
it is doable with nrf52, we are doing it with cdc interface. it could upgrade itself via cdc packet. Maybe it is not too difficult to get it work with uf2

hathach:
bootloader + SD uf2 image will have an different uf2 family ID to prevent accident or corrupted image from normal application uf2.

PR #88, opened 2019-11-29
adding support for nrf52833

hathach:
nrf52833 complied but couldn't run just yet. Since S140 v6 does not support SOC on 833. We will need to upgrade S140 to v7 #72 for 52833. Though this PR refactor bootloader, and pave the way to support  5340 later on. There should be no behavior changes to current 52832 and 52840.

we will wrap it up for now and come back to add 833 support after upgrade #72 and #24


hathach:
thanks @dhalbert, this PR still works with all existing boars + SD v6 as previously. Does upgrade nrfx to v2 and compile with 833, although 833's produced binary is not usable. It is good to merge now.

dhalbert:
OK, good, I can start to work on the nrfx V2 soon!

hathach:
@dhalbert I am currently working on the v2 upgrade, just wait for it :)

Issue #74, opened 2019-08-04
Use without softdevice


antst:
Hello,

I am trying to adopt this bootloader for use in non-BLE application, without softdevice.
Mainly due to UF2. 

By commenting out -DBLE_STACK_SUPPORT_REQD and -DSOFTDEVICE_PRESENT and excluding compilation of few BLE-specific binaries and wrapping in ifdef few BLE-specific calls I was able to compile it, but it doesn't work for me. (I test on nrf52840-DK with flashed MBR)
Looking that there are those defines, it looks like somebody had in mind use of this boot-loader in non-BLE application. 
Were there attempts to run without softdevice? Any suggestions?

Thanks!

xiongyihui:
+1
want to use it with zephyr or mynewt which has its own ble stack

hathach:
related to #30

hathach:
implemented by #128

Issue #30, opened 2018-09-07
Bootloader uf2/cdc should run even without SD flashed


hathach:
improve stability by making the bootloader uf2/cdc running even without SD. The work is already 90% done, the rest is setting that bootloader pull from the SD, MBR etc ... e.g SD size to flash to appropriate address etc ... it should use default one ( e.g SD 6.1.0 known  size in case SD is not flashed). 

Once the bootloader without SD (or corrupted) can be put into known good by **dfu-flash** with an pre-built binary in the bin directory.

hathach:
closed, we no longer want to do so. Could reopen if needed later.

PR #108, opened 2020-02-03
a step to no softdevice

xiongyihui:
Make the bootloader work without softdevice.

comment out https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/master/Makefile#L245 to enable it.

```
# CFLAGS += -DSOFTDEVICE_PRESENT
```

The MBR at 0x0 to 0x1000 is still required.


hathach:
related to #30  #74

hathach:
thanks for your PR, however, we haven't decided if we want to have SD-independent bootloader yet. Meanwhile this PR will be pending until we make up our mind.

jrobeson:
@hathach : newer sdks include almost all of this PR out of the box, so SD free bootloader is almost free code wise (but perhaps not support wise)

hathach:
@jrobeson since we have plan to extensive use OTA in the future which requires SD. We don't want an incompatible  bootloader code and Softdevice (e.g v6, v7). And then we have to document which bootloader version work with which Softdevice. Worse an incompatible Softdevice (flashed from previously compatible) can potentially brick device. So we like to keep it as a combo for now, until there is a real reason to separate them.

PR #110, opened 2020-02-14
CF2 support

### LINKED ISSUE #92: CF2 support

UF2 defines a format to store configuration data about the board in the bootloader area, see https://github.com/microsoft/uf2/blob/master/cf2.md

This is currently used primarily by MakeCode Arcade to define where the button/screen pins are, but is extensible. This is supported in SAMD51 and STM32F4 bootloaders.

Could we have maybe 1k of bootloader area reserved for this?

mmoskal:
This adds a 1kB section at the end of the bootloader for [CF2 configuration data](https://github.com/microsoft/uf2/blob/master/cf2.md). I also add an example Arcade board configuration based on Feather. In fact, we could define it for all boards, that would make supporting them in pxt-maker slightly easier.

Added support in PXT - https://github.com/microsoft/pxt-common-packages/pull/1074

Fixes #92 

Additionally:
* add support for flashing with BMP and GDB
* add support for booting straight to app without waiting

I may add support for single-press reset (optional, based on request by user app), as implemented in SAMD21/51.

### DIFF
--- a/src/boards/arcade_feather_nrf52840_express/board.h
+++ b/src/boards/arcade_feather_nrf52840_express/board.h
@@ -1,0 +1,143 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2018 Ha Thach for Adafruit Industries
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#ifndef _FEATHER_NRF52840_H
+#define _FEATHER_NRF52840_H
+
+#define _PINNUM(port, pin)    ((port)*32 + (pin))
+
+/*------------------------------------------------------------------*/
+/* LED
+ *------------------------------------------------------------------*/
+#define LEDS_NUMBER           2
+#define LED_PRIMARY_PIN       _PINNUM(1, 15)
+#define LED_SECONDARY_PIN     _PINNUM(1, 10)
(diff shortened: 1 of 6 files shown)


mmoskal:
86180244_676077103129929_574737216287277056_n
86295576_601529660699537_5367009670160449536_n

CC @ladyada

ladyada:
!@!@ im thinkin' of making an nRF52840 pybadge - now i have something to test with :)

mmoskal:
The last commit should allow for using reset button essentially to swap between application and bootloader. This is controlled by a flag in the app and is the behavior of CPX and all Arcade boards when using MakeCode.

src/boards/arcade_feather_nrf52840_express/board.h
```diff
@@ -0,0 +1,143 @@
...
+//------------- UF2 -------------//
+#define UF2_PRODUCT_NAME      "Adafruit Feather nRF52840 Express"
+#define UF2_VOLUME_LABEL      "ARCADE-N4"
+#define UF2_BOARD_ID          "nRF52840-Feather-revD"
+#define UF2_INDEX_URL         "https://www.adafruit.com/product/4062"
+
+#ifdef DEFINE_BOOTLOADER_CONFIG
+#define BOOTLOADER_CONFIG_PRESENT 1
+__attribute__((used, section(".bootloaderConfig")))
+const uint32_t bootloaderConfig[] = {
```
hathach: Thanks, I understand the great idea behind CF2 now :) . Though 1KB is only 128 (minus 1) pairs of config. It is a bit small IMHO.  Should we go to 2KB for 512 pairs. The current head room for bootloader is still plenty for expanding. 

Total size is 40KB, current compiled is 30KB, 2 KB for CF2, and 8KB left is good enough

src/linker/nrf52840_s140_v6.ld
```diff
@@ -11,7 +11,9 @@ MEMORY
    *  those values do not match. The check is performed in main.c, see
    *  APP_ERROR_CHECK_BOOL(*((uint32_t *)NRF_UICR_BOOT_START_ADDRESS) == BOOTLOADER_REGION_START);
    */
-  FLASH (rx) : ORIGIN = 0xF4000, LENGTH = 0xFE000-0xF4000 /* 40 KB */
+  FLASH (rx) : ORIGIN = 0xF4000, LENGTH = 0xFE000-0xF4000-1024 /* 39 KB */
+
+  BOOTLOADER_CONFIG (r): ORIGIN = 0xFE000 - 1024, LENGTH = 1024
```
hathach: we could increase it to 2KB leaving a total of 38KB for bootloader, which is quite enough for exanding.

src/boards/arcade_feather_nrf52840_express/board.h
```diff
@@ -0,0 +1,143 @@
...
+
+//------------- UF2 -------------//
+#define UF2_PRODUCT_NAME      "Adafruit Feather nRF52840 Express"
+#define UF2_VOLUME_LABEL      "ARCADE-N4"
+#define UF2_BOARD_ID          "nRF52840-Feather-revD"
+#define UF2_INDEX_URL         "https://www.adafruit.com/product/4062"
+
+#ifdef DEFINE_BOOTLOADER_CONFIG
+#define BOOTLOADER_CONFIG_PRESENT 1
+__attribute__((used, section(".bootloaderConfig")))
```
hathach: @mmoskal I just notice this is header file, we shouldn't declare variable here, it is probably better to define it as macro to series of values and put the variable in a C file (main.c or uf2.c etc ..) . I did something similar to my stack with ascii to hid keycode table here 

https://github.com/hathach/tinyusb/blob/master/src/class/hid/hid.h#L617

hathach: I made an PR to move bootloader config to C file https://github.com/adafruit/Adafruit_nRF52_Bootloader/pull/111 , please check it out

src/boards/arcade_feather_nrf52840_express/board.h
mmoskal: sounds good, will do!

PR #111, opened 2020-02-28
move bootloaderConfig into pinconfig.c

hathach:
This is PR to PR, I still don't feel right to declare variable inside header file even with guard.


mmoskal:
This looks good! I'll need to update https://microsoft.github.io/uf2/patcher/ to handle this text format but no big deal.

hathach:
@mmoskal oops, Sorry, I am not aware that you have to change the patcher for this, maybe we should have a `arcade_feather_nrf52840_express.c` that actually declare the array in normal C format.

PR #112, opened 2020-02-29
follow up to pincofig

hathach:
@mmoskal I think it is easier for the `patcher`  to generate an complete C file than MACRO define.  User can put it into the board folder which is optional. Sorry for changing my mind, I wasn't aware that you have to change your script. Hope this makes thing easier.


mmoskal:
Thanks!

hathach:
@mmoskal Could I ask how can I test this CF2 with makecode, I haven't work with makecode before. I am trying to add CF2 for CLUE as well, since it has screen and button which could be good candidate.

mmoskal:
I should be able to deploy this next week in Maker, and soon after that in Arcade. As for the CLUE, not sure it will work in Arcade, as the screen size doesn't match, but it should work in Maker soon.

PR #114, opened 2020-03-10
Make sure all pins are in reset state on teardown

mmoskal:
In particular, the button pins are left configured with sense, which may (and did in my case) wreak havoc in the application.

### DIFF
--- a/src/boards.c
+++ b/src/boards.c
@@ -115,6 +115,13 @@ void board_teardown(void)
 
   // Stop LF clock
   NRF_CLOCK->TASKS_LFCLKSTOP = 1UL;
+
+  // make sure all pins are back in reset state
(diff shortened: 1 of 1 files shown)


src/boards.c
```diff
@@ -115,6 +115,15 @@ void board_teardown(void)
 
   // Stop LF clock
   NRF_CLOCK->TASKS_LFCLKSTOP = 1UL;
+
+  // make sure all pins are back in reset state
+  for (int i = 0; i < 32; ++i)
```
hathach: great catch, though I think we should use `nrf_gpio_cfg_default()`, which is designed to do so. 

https://github.com/NordicSemiconductor/nrfx/blob/master/hal/nrf_gpio.h#L222

PR #124, opened 2020-04-30
S340 for nrf52840 feather

orrmany:
Hi, 

In connection to [PR #494 for Adafruit_nRF52_Arduino](https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/494) here comes the PR for the bootloader. 

I would like to ask for some modifications to the nRF52 bootloader  for making it easier to use nRF52840 Express with BLE & ANT+ protocol stacks together. That necessitates modificiations to the bootloader to have  the S340 BLE & ANT multiprotocol SoftDevice (SD) instead of the stock S140 SD.
(which will come in a separate PR, since that is a different repo). The modification leaves S140 in place, since instead of replacing S140, it rather defines a new board type "feather_nrf52840_express_s340" for S340. 

I believe that these changes together with [PR #494 for Adafruit_nRF52_Arduino](https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/494) makes possible to reuse _all_ Bluefruit52 code as-is with S340 _and_ enables ANT+ communication. 

Of course, in order to be able to build the firmware for S340 one needs to register at thisisant.com, obtain the license and download the softdevice.  A readme.txt has been added to the repo describing where and how shall the user insert the S340 header and hex files into the source tree. 

You can read further details at my Wiki below (Note, those articles were written a few months ago, so they reflect an earlier SW version of your code): 
- [S340 SoftDevice & Adafruit nRF52840 Feather Express](http://orrmany.hu/wiki/doku.php?id=arduino:s340_softdevice_adafruit_nrf52840_feather_express)
- [Arduino IDE integration for the nRF52840 Feather Express with S340](http://orrmany.hu/wiki/doku.php?id=s340:arduino_ide_integration_for_the_nrf52840_feather_express_with_s340)
- [PlatformIO IDE integration for the nRF52840 Feather Express with S340](http://orrmany.hu/wiki/doku.php?id=s340:platformio_ide_integration_for_the_nrf52840_feather_express_with_s340)

Credit shall also go to @rtgree01 ([Ryan Green](https://github.com/rtgree01/Adafruit_nRF52_Arduino)) who made most of the early hacks of these and he kindly shared his results with me and @cujomalainey [Curtis Malainey](https://github.com/cujomalainey).

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -326,10 +326,10 @@ static uint32_t softdev_init(bool init_softdevice)
       .accuracy     = NRF_CLOCK_LF_ACCURACY_250_PPM
   };
 
-  #ifndef ANT_LICENSE_KEY
-  APP_ERROR_CHECK( sd_softdevice_enable(&clock_cfg, app_error_fault_handler) );
-  #else
+  #ifdef ANT_LICENSE_KEY
   APP_ERROR_CHECK( sd_softdevice_enable(&clock_cfg, app_error_fault_handler, ANT_LICENSE_KEY) );
+  #else
+  APP_ERROR_CHECK( sd_softdevice_enable(&clock_cfg, app_error_fault_handler) );
   #endif
 
   sd_nvic_EnableIRQ(SD_EVT_IRQn);
--- a/src/main.c
+++ b/src/main.c
@@ -326,7 +326,12 @@ static uint32_t softdev_init(bool init_softdevice)
       .accuracy     = NRF_CLOCK_LF_ACCURACY_250_PPM
   };
 
+  #ifndef ANT_LICENSE_KEY
   APP_ERROR_CHECK( sd_softdevice_enable(&clock_cfg, app_error_fault_handler) );
+  #else
+  APP_ERROR_CHECK( sd_softdevice_enable(&clock_cfg, app_error_fault_handler, ANT_LICENSE_KEY) );
+  #endif
+
   sd_nvic_EnableIRQ(SD_EVT_IRQn);
 
   /*------------- Configure BLE params  -------------*/
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,6 @@ bin/
 *.jlink
 
 TAGS
+# Ignore S340 SoftDevice due to license restrictions
+bootloader/feather_nrf52840_express/feather_nrf52840_express_bootloader-0.3.0_s340_6.1.1.*
+cores/nRF5/nordic/softdevice/s340_nrf52_6.1.1_API/*
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ SD_HEX       = $(SD_PATH)/$(SD_FILENAME)_softdevice.hex
 LD_FILE      = linker/$(MCU_SUB_VARIANT)_$(SD_NAME)_v$(word 1, $(subst ., ,$(SD_VERSION))).ld
 
 GIT_VERSION = $(shell git describe --dirty --always --tags)
-GIT_SUBMODULE_VERSIONS = $(shell git submodule status | cut -d' ' -f3,4 | paste -s -d" " -)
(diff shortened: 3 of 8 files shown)


orrmany:
Now, this is interesting: the CI check step 'build' for "feather_nrf52840_express_s340" has been failed, since I have not added the license-bound S340 files. As such, this build-failure is "intentious". Advice how to handle this situation would be highly appreciated.

src/main.c
```diff
@@ -326,7 +326,12 @@ static uint32_t softdev_init(bool init_softdevice)
       .accuracy     = NRF_CLOCK_LF_ACCURACY_250_PPM
   };
 
+  #ifndef ANT_LICENSE_KEY
```
cujomalainey, line 329: maybe we should depend on USE_S340 here so we don't call the wrong function? also minor nit:no need to negate, best to flip cases for readibility.

orrmany: Flipping is no problem. The condition is better for ANT_LICENSE_KEY, IMHO, since the new code actively uses ANT_LICENSE_KEY. Switching to USE_340 would solve one problem but would create a similar one: the risk of calling the method with undefined symbol. Moreover, this a C-linkage function, passing an extra parameter (ANT_LICENSE_KEY) to the S140 variant of the method would make no harm: the extra  arg. would just silently sit on the heap, without any harm.

orrmany: Actually  the code ".source       = NRF_CLOCK_LF_SRC_RC", that is, the usage of the RC oscillator instead of the quartz is the big-big problem here, since the low-accuracy RC oscillator won't fly for ANT :/

orrmany:
@hathach: any advice how to solve this "we need a skeleton board" situation here?

src/main.c
orrmany: Commit e3683b3 makes the requested flip

hathach:
Thank you for your PR, we are currently decoupling the Softdevice from bootloader (still in the making) https://github.com/adafruit/Adafruit_nRF52_Bootloader/issues/126
It will allow you or any other users to replace softdevice with any other stack of choice e.g S340. Therefore this PR wouldn't be needed for you to add S340 support to existing board. Though I will leave it open until we got the new release out.

orrmany:
Hi @hathach , I am not sure if I understand the USB lingo in that issue, I am sorry.
Basically this is my use case:

Suppose that I, as an Adafruit customer, previously purchased an Adafruit MCU with an nRF52 SoC onboard.

Now, also suppose that I, as an Arduino developer, would like to be able replace the stock SoftDevice (SD) delivered with product with a multiprotocol SD alternative, if such one exists. nRF52832, or nRF52840 are multiprotocol capable, for example.

    E.g., I would like to have the S140 SD of the nRF52840 Feather Express be replaced with S340.
    Another example would be an nRF52832 customer, who would like to replace his S132 SD with a multiprtocol S332.
    etc.

I would like to retain all legacy Adafruit functionality with the alternative, multiprotocol SoftDevices, since the functionality of the alternative SD-s are extensions of that of their lesser (BLE only) counterparts. That is, I would like to use CircuitPython, DFU, Bluefruit, etc. with the multiprotocol SD-s.

Could you explain what of those checklist items of yours in issue #126  are applicable to my use case?

Thanks, Gábor

hathach:
checkilist is test list, 
- cdc interface will work as it is right now. 
- UF2 currently only allows updating application, it will allow update application + softdevice + bootloader in later version. You can use it flash S340 (in form of uf2) to stock bootloader without the need of jlink

orrmany:
Ok. Btw: I only used jlink for unbricking occasionally. The modification
here worked with plain serial DFU, as far as I remember.

Ha Thach <notifications@github.com> ezt írta (időpont: 2020. máj. 6., Sze,
19:48):

hathach:
closed, since we don't have hex, specs or any information to maintain the S340. In fact, we have no plan to maintain other SoftDevice than S140 and S132.

PR #127, opened 2020-05-09
Add note about installing Git LFS

zjwhitehead:
Sub-modules in this project need Git LFS support in order to properly clone


hathach:
Yeah, it is due to the cmsis 5, however despite what the readme says, we don’t need to clone submodules of tinyusb to build the bootloader. Therefore this pr is not necessary, I will correct readme.

PR #132, opened 2020-05-26
Enable large volume size

henrygab:
# Enable large (~31MB) volume size in GhostFAT

This also includes a fix for #133 via 3b92ef8.

I can easily open a PR on master later, no problem.

### DIFF
--- a/src/usb/uf2/ghostfat.c
+++ b/src/usb/uf2/ghostfat.c
@@ -300,11 +300,11 @@ void read_block(uint32_t block_no, uint8_t *data) {
             d->createTime       = __DOSTIME__;
             d->createDate       = __DOSDATE__;
             d->lastAccessDate   = __DOSDATE__;
-            d->highStartCluster = startCluster >> 8; // BUGBUG -- shouldn't this be 16?
+            d->highStartCluster = startCluster >> 16;
             // DIR_WrtTime and DIR_WrtDate must be supported
             d->updateTime       = __DOSTIME__;
             d->updateDate       = __DOSDATE__;
-            d->startCluster     = startCluster & 0xFF; // BUGBUG -- shouldn't this be 0xFFFF?
+            d->startCluster     = startCluster & 0xFFFF;
             d->size = (inf->content ? strlen(inf->content) : UF2_SIZE);
         }
 
--- a/src/usb/uf2/ghostfat.c
+++ b/src/usb/uf2/ghostfat.c
@@ -86,6 +86,7 @@ STATIC_ASSERT(sizeof(DirEntry)                             ==        32); // FAT
 STATIC_ASSERT(BPB_SECTOR_SIZE % sizeof(DirEntry)           ==         0); // FAT requirement
 STATIC_ASSERT(BPB_ROOT_DIR_ENTRIES % DIRENTRIES_PER_SECTOR ==         0); // FAT requirement
 STATIC_ASSERT(BPB_SECTOR_SIZE * BPB_SECTORS_PER_CLUSTER    <= (32*1024)); // FAT requirement (64k+ has known compatibility problems)
+STATIC_ASSERT(FAT_ENTRIES_PER_SECTOR                       ==       256); // FAT requirement
 
 #define STR0(x) #x
 #define STR(x) STR0(x)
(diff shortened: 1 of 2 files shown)


henrygab:
@hathach -- OK, this should work.  I used `STATIC_ASSERT()` to help validate at compilation time that the file system generated would be valid (where I could).

The final commit also sets the total number of sectors to a fixed value.  There's still a static assert that, if `CFG_UF2_NUM_BLOCKS` is defined, it's less than or equal to the value GhostFAT is now hard-coded to use.

Please feel free to send any questions my way.

hathach:
superb!! Though we need to use `CFG_UF2_NUM_BLOCKS = BPB_TOTAL_SECTORS` or since it is used to response to the SCSI Read Capacity 

image

src/usb/uf2/ghostfat.c
```diff
@@ -63,7 +63,33 @@ struct TextFile {
...
+#define FAT_ENTRY_SIZE            (2)
+#define FAT_ENTRIES_PER_SECTOR    (BPB_SECTOR_SIZE / FAT_ENTRY_SIZE)
+// NOTE: MS specification explicitly allows FAT to be larger than necessary
+#define BPB_SECTORS_PER_FAT       ( (BPB_TOTAL_SECTORS / FAT_ENTRIES_PER_SECTOR) + \
+                                   ((BPB_TOTAL_SECTORS % FAT_ENTRIES_PER_SECTOR) ? 1 : 0))
+#define DIRENTRIES_PER_SECTOR     (BPB_SECTOR_SIZE/sizeof(DirEntry))
+#define ROOT_DIR_SECTOR_COUNT     (BPB_ROOT_DIR_ENTRIES/DIRENTRIES_PER_SECTOR)
+
+#if defined(CFG_UF2_NUM_BLOCKS)
+  STATIC_ASSERT(BPB_TOTAL_SECTORS >= CFG_UF2_NUM_BLOCKS); // configuration used, and asked for a larger size GhostFAT?
```
hathach, line 82: we can just used `CFG_UF2_NUM_BLOCKS` as BPB_TOTAL_SECTORS or vice versa (1 as alias to other). Since the value is used by usb_msc to response to SCSI Read Capacity command.

src/usb/uf2/ghostfat.c
```diff
@@ -270,35 +315,37 @@ void read_block(uint32_t block_no, uint8_t *data) {
             d->createTime       = __DOSTIME__;
             d->createDate       = __DOSDATE__;
             d->lastAccessDate   = __DOSDATE__;
-            d->highStartCluster = startCluster >> 8;
+            d->highStartCluster = startCluster >> 16;
```
hathach, line 318: ah I see the issue #133  now :+1:

src/usb/uf2/ghostfat.c
```diff
@@ -518,5 +565,6 @@ int write_block (uint32_t block_no, uint8_t *data, WriteState *state)
     }
   }
 
-  return 512;
+  STATIC_ASSERT(BPB_SECTOR_SIZE == 512); // if sector size changes, may need to re-validate this code
```
hathach, line 568: despite of whatever Microsoft says, sector never CHANGES :laughing: :laughing:

src/usb/uf2/ghostfat.c
```diff
@@ -63,7 +63,33 @@ struct TextFile {
...
+#define BPB_SECTORS_PER_FAT       ( (BPB_TOTAL_SECTORS / FAT_ENTRIES_PER_SECTOR) + \
+                                   ((BPB_TOTAL_SECTORS % FAT_ENTRIES_PER_SECTOR) ? 1 : 0))
+#define DIRENTRIES_PER_SECTOR     (BPB_SECTOR_SIZE/sizeof(DirEntry))
+#define ROOT_DIR_SECTOR_COUNT     (BPB_ROOT_DIR_ENTRIES/DIRENTRIES_PER_SECTOR)
+
+#if defined(CFG_UF2_NUM_BLOCKS)
+  STATIC_ASSERT(BPB_TOTAL_SECTORS >= CFG_UF2_NUM_BLOCKS); // configuration used, and asked for a larger size GhostFAT?
+#endif
+
+STATIC_ASSERT(BPB_SECTOR_SIZE                              ==       512); // GhostFAT does not support other sector sizes (currently)
```
hathach, line 85: I normally put a warnings near these macro definition and threaten people not to change it. But this works as well :smile:

src/usb/uf2/ghostfat.c
henrygab: Thank you for catching that!  I did not know it was used elsewhere.  👍

src/usb/uf2/ghostfat.c
henrygab: For FAT16, you are right.

hathach: nobody will change this value even with FAT32, probably only existed in other place than embedded :)

PR #136, opened 2020-05-29
Self uf2 large volume

### LINKED ISSUE #129: Increase NUM_FAT_BLOCKS to maximum as possible
**Is your feature request related to a problem? Please describe.**
Currently NUM_FAT_BLOCKS is hard-coded to 16MB.  (0x8000 sectors)
Is there any reason this is not set to a much larger number?

One reason to use a large (virtual) volume size:
* The minimum size needed for the UF2 is more than 2x the flash size (current.uf2, metadata, fs overhead, ...)
* Some OS are "smart" about free size, and also remove the size of their metadata writes from free size, further reducing available space for upload
* One goal of UF2 was to allow concatenation of multiple UF2 files (with different ~~board IDs~~ family), so that the same UF2 file could be copied to many different devices, and each would only write the data for its own ~~board id~~ family.  However, that large file would fail to copy if there is insufficient (virtual) free space on the media.

_Originally posted by @henrygab in https://github.com/adafruit/Adafruit_nRF52_Bootloader/pull/128_

Increase the UF2 drive size to maximum within the specs and/or memory allowed by nRF52

### LINKED ISSUE #133: Incorrect cluster calculations
https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/79a6a0c4f7f526704d5b47ffd4f6020265e01444/src/usb/uf2/ghostfat.c#L253-L257

Both `highStartCluster` and `startCluster1` are of type `uint16_t`.  Therefore, the math here is wrong, and the wrong starting cluster would be generated for any file that starts at sector 0x100 or higher.

Easy enough to fix.  I'll include in my PR to support larger volumes.

henrygab:
# Enable large (~31MB) volume size in GhostFAT

See original discussion at PR #132.

@hathach  -- this should look very familiar.  I just re-based, and added [28e09e0](https://github.com/adafruit/Adafruit_nRF52_Bootloader/commit/28e09e0abaa287ac47b611fd579bc1d411f67556) to fix the `READ_CAPACITY` issue.

This also includes a fix for #133 via [2c9826a](https://github.com/adafruit/Adafruit_nRF52_Bootloader/commit/2c9826acd9aef57e0dcff188e8217c9f3d66468a).

Fixes #133
Fixes #129

### DIFF
--- a/src/usb/uf2/ghostfat.c
+++ b/src/usb/uf2/ghostfat.c
@@ -300,11 +300,11 @@ void read_block(uint32_t block_no, uint8_t *data) {
             d->createTime       = __DOSTIME__;
             d->createDate       = __DOSDATE__;
             d->lastAccessDate   = __DOSDATE__;
-            d->highStartCluster = startCluster >> 8; // BUGBUG -- shouldn't this be 16?
+            d->highStartCluster = startCluster >> 16;
             // DIR_WrtTime and DIR_WrtDate must be supported
             d->updateTime       = __DOSTIME__;
             d->updateDate       = __DOSDATE__;
-            d->startCluster     = startCluster & 0xFF; // BUGBUG -- shouldn't this be 0xFFFF?
+            d->startCluster     = startCluster & 0xFFFF;
             d->size = (inf->content ? strlen(inf->content) : UF2_SIZE);
         }
 
--- a/src/usb/uf2/ghostfat.c
+++ b/src/usb/uf2/ghostfat.c
@@ -86,6 +86,7 @@ STATIC_ASSERT(sizeof(DirEntry)                             ==        32); // FAT
 STATIC_ASSERT(BPB_SECTOR_SIZE % sizeof(DirEntry)           ==         0); // FAT requirement
 STATIC_ASSERT(BPB_ROOT_DIR_ENTRIES % DIRENTRIES_PER_SECTOR ==         0); // FAT requirement
 STATIC_ASSERT(BPB_SECTOR_SIZE * BPB_SECTORS_PER_CLUSTER    <= (32*1024)); // FAT requirement (64k+ has known compatibility problems)
+STATIC_ASSERT(FAT_ENTRIES_PER_SECTOR                       ==       256); // FAT requirement
 
 #define STR0(x) #x
 #define STR(x) STR0(x)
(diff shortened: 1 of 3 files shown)


hathach:
sorry for the trouble, I didn't mean to close #132 , it is closed by github automatically since the `self-uf2` branch is removed when merged into master.

src/usb/uf2/ghostfat.c
```diff
@@ -90,50 +112,62 @@ static struct TextFile const info[] = {
...
-#define UF2_FIRST_SECTOR   (NUM_FILES + 1) // WARNING -- code presumes each non-UF2 file content fits in single sector
-#define UF2_LAST_SECTOR    (UF2_FIRST_SECTOR + UF2_SECTORS - 1)
+#define UF2_FIRMWARE_BYTES_PER_SECTOR 256
+#define TRUE_USER_FLASH_SIZE (USER_FLASH_END-USER_FLASH_START)
+STATIC_ASSERT(TRUE_USER_FLASH_SIZE % UF2_FIRMWARE_BYTES_PER_SECTOR == 0); // UF2 requirement -- overall size must be integral multiple of per-sector payload?
 
-#define RESERVED_SECTORS   1
-#define ROOT_DIR_SECTORS   4
-#define SECTORS_PER_FAT    ((NUM_FAT_BLOCKS * 2 + 511) / 512)
+#define UF2_SECTORS        ( (TRUE_USER_FLASH_SIZE / UF2_FIRMWARE_BYTES_PER_SECTOR) + \
```
hathach, line 142: the above static check already ensure it is evenly divided by UF2_FIRMWARE_BYTES_PER_SECTOR, there is no need for this computation. The 256 bytes per sector won't be changed, the flash range is only changes with 4KB page as well. We can just use it the way it is as before. Let's simplify these macro, possibly remove the `TRUE_USER_FLASH_SIZE`

src/usb/uf2/ghostfat.c
```diff
@@ -63,7 +63,29 @@ struct TextFile {
...
+// NOTE: MS specification explicitly allows FAT to be larger than necessary
+#define BPB_SECTORS_PER_FAT       ( (BPB_TOTAL_SECTORS / FAT_ENTRIES_PER_SECTOR) + \
+                                   ((BPB_TOTAL_SECTORS % FAT_ENTRIES_PER_SECTOR) ? 1 : 0))
+#define DIRENTRIES_PER_SECTOR     (BPB_SECTOR_SIZE/sizeof(DirEntry))
+#define ROOT_DIR_SECTOR_COUNT     (BPB_ROOT_DIR_ENTRIES/DIRENTRIES_PER_SECTOR)
+
+STATIC_ASSERT(BPB_SECTOR_SIZE                              ==       512); // GhostFAT does not support other sector sizes (currently)
+STATIC_ASSERT(BPB_SECTORS_PER_CLUSTER                      ==         1); // GhostFAT presumes one sector == one cluster (for simplicity)
+STATIC_ASSERT(BPB_NUMBER_OF_FATS                           ==         2); // FAT highest compatibility
+STATIC_ASSERT(sizeof(DirEntry)                             ==        32); // FAT requirement
```
hathach, line 84: this is already checked after the DirEntry definition.. Look at this now, I think we don't really need to perform these static assert, just put an warning before these BPB and tell user not to change it. Basically we static check thing that could be wrong like the `__attribute__((packed)) DirEntry;` not the actual macro with fixed magic number

henrygab: static assertions catch user error.
user error can include accidental and unintentional changes (as in your example).

user ignorance can also cause a user error.
a later editor (not you) might change something without full understanding.

therefore, static assert also is useful for making explicit code assumptions.

This code had many assumptions.

src/usb/uf2/ghostfat.c
henrygab: If you feel strongly, you may change it.
I would not recommend it.

I feel it is better (especially when zero code impact) to **_code defensively_**, which means to catch errors at the earliest possible time.  Every later stage an issue is found increases impact and time to fix:
* Edit  >>  Compile  >> Link  >>  Pre-execution   >>  Run-time

You have ability to edit this branch directly, so I will not stop you if you feel so strongly.

src/usb/uf2/ghostfat.c
```diff
@@ -518,5 +561,6 @@ int write_block (uint32_t block_no, uint8_t *data, WriteState *state)
     }
   }
 
-  return 512;
+  STATIC_ASSERT(BPB_SECTOR_SIZE == 512); // if sector size changes, may need to re-validate this code
```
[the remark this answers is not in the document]
henrygab: You are right, that it exists elsewhere.
I had it here because I am unsure if the caller expects 512, or if the caller can accept 4k result.

The English saying, "belt and suspenders" applies... you could always wear only one, but wearing both belt and suspenders makes it even less likely your pants will fall.

Again, if you feel strongly, you can edit this branch directly before accepting this PR.  I will not stop you.

However, again because it has zero impact on final code, it is my opinion that the cost of an extra line of text is worthwhile.

src/usb/uf2/ghostfat.c
hathach: I agree with you that we should catch error asap, however, static check right after the macro define with the same value is not catching bug IMHO. It is more on the code duplication.

src/usb/uf2/ghostfat.c
hathach: we don't need to have static check on every magic value, those magic constant is part of any C source.

hathach:
Thanks @henrygab for another great PR :smiley:  It will definitely take me much longer to do something like this. We are getting close to the next big release for bootloader :)

src/usb/uf2/ghostfat.c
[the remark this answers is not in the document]
hathach: Yeah, I feel strongly, but I am even more lazy. I will leave it as it is now and will **clean up** later on when I work with bootloader.

PR #137, opened 2020-06-05
Add nice!nano development board

Nicell:
I've added the configuration for my nRF52840 based development board I've created. It will be available soon. There's some bare bones documentation available here that will get extended soon: https://docs.nicekeyboards.com/#/nice!nano/

Thanks for the work that's been put into this bootloader! It's been a joy to use.

### DIFF
--- a/src/boards/nice_nano/board.h
+++ b/src/boards/nice_nano/board.h
@@ -1,0 +1,63 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 Nick Winans
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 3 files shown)


src/boards/nice_nano/board.h
```diff
@@ -0,0 +1,63 @@
...
+// BLE OTA
+//--------------------------------------------------------------------+
+#define BLEDIS_MANUFACTURER  "Nice Keyboards"
+#define BLEDIS_MODEL         "nice!nano"
+
+//--------------------------------------------------------------------+
+// USB
+//--------------------------------------------------------------------+
+#define USB_DESC_VID           0x239A
+#define USB_DESC_UF2_PID       0x0029
```
hathach, line 55: the VID/PID is already used and we don't allow to reuse it anymore. If your board can also run circuitpython, you could apply for a set of VID/PID from Adafruit.

Nicell: I'm not familiar with Circuit Python. Is there any reason I wouldn't be able to use Circuit Python on this nRF52840 board? If not, could you direct me where to add support and apply for a VID/PID from Adafruit? Thanks!

hathach: join circuitpyhon discords channel and ask for a donation of VID/PID to see if you are eligible for one.

Nicell:
@hathach I've added my unique PID. See: https://github.com/adafruit/circuitpython/issues/3013

PR #140, opened 2020-06-15
Change SEGGER options to non blocking print.

geeksville:
In the previous config if you were printing a lot in your bootloader
debug and you didn't have an ICE connected, the boot would stall waiting
for the SEGGER to read the print buffer.  Better I think to just throw
away the oldest contents of the ring buffer if the ICE is not connected.


geeksville:
though this shouldn't be related to segger dropping messages, because their ring buffer is fairly large.  This change just changes the behavior on what happens when that ring-buffer fills up.  When that buffer is full it seems bad to me to stall the boot (because if the user is not debugging their bootloader seemingly hangs)

hathach:
The problem is, without the blocking wait, the message lost when I purposely watching the log with RTT. Which will swallow the context and information needed for troubleshooting which is the reason why the DEBUG with that log is intended for. In short, when RTT is used, I preferr using it with a full info. That's why I suggest to increase to the DEBUG=1 to DEBUG=2 to keep compatible with previous `DEBUG=1`  behavior to only build with -ggdb . If you want to run without debugger attached, please use the lesser DEBUG level

geeksville:
no problem!

PR #143, opened 2020-07-08
Reset into app, if no USB connection in 3s

mmoskal:
This only applies in SINGLE_TAP_RESET mode (MakeCode) - in that mode the board goes into bootloader by default for easy programming. However, if the board is not connected to computer, we should go into app (this is what we do on SAMD and STM32). This sets a 3000ms timeout if the USB is not enumerated.

Note, that such long delay was needed for certain versions of macOS - 10.14.4 until, but not including, Catalina.

### DIFF
--- a/lib/sdk11/components/libraries/bootloader_dfu/bootloader.c
+++ b/lib/sdk11/components/libraries/bootloader_dfu/bootloader.c
@@ -49,8 +49,7 @@ typedef enum
 
 static pstorage_handle_t        m_bootsettings_handle;  /**< Pstorage handle to use for registration and identifying the bootloader module on subsequent calls to the pstorage module for load and store of bootloader setting in flash. */
 static bootloader_status_t      m_update_status;        /**< Current update status for the bootloader module to ensure correct behaviour when updating settings and when update completes. */
-
-static bool m_cancel_timeout_on_usb;
+static bool m_cancel_timeout_on_usb; /**< If set the timeout is cancelled when USB is enumerated. Otherwise, the timeout is only cancelled when DFU update is started. */
 
 APP_TIMER_DEF( _dfu_startup_timer );
 volatile bool dfu_startup_packet_received = false;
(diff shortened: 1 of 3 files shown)


lib/sdk11/components/libraries/bootloader_dfu/bootloader.c
```diff
@@ -313,10 +322,12 @@ uint32_t bootloader_init(void)
 }
 
 
-uint32_t bootloader_dfu_start(bool ota, uint32_t timeout_ms)
+uint32_t bootloader_dfu_start(bool ota, uint32_t timeout_ms, bool cancel_timeout_on_usb)
 {
     uint32_t err_code;
 
+    m_cancel_timeout_on_usb = cancel_timeout_on_usb;
```
hathach, line 329: This is a great idea, though I think we should exclude the OTA mode from this timeout. Although we didn't fully support OTA mode, it is still doable via nRFConnect app. 

```
m_cancel_timeout_on_usb = cancel_timeout_on_usb && !ota
```

lib/sdk11/components/libraries/bootloader_dfu/bootloader.c
```diff
@@ -50,6 +50,8 @@ typedef enum
 static pstorage_handle_t        m_bootsettings_handle;  /**< Pstorage handle to use for registration and identifying the bootloader module on subsequent calls to the pstorage module for load and store of bootloader setting in flash. */
 static bootloader_status_t      m_update_status;        /**< Current update status for the bootloader module to ensure correct behaviour when updating settings and when update completes. */
 
+static bool m_cancel_timeout_on_usb;
```
hathach, line 53: Please put a few text describing the variable just in case we forgot later on e.g

```
// Should we jump to application if USB is not mounted in specificed interval
```

src/main.c
```diff
@@ -250,7 +250,14 @@ int main(void)
     }
 
     // Initiate an update of the firmware.
-    APP_ERROR_CHECK( bootloader_dfu_start(_ota_dfu, 0) );
+    if (APP_ASKS_FOR_SINGLE_TAP_RESET())
+    {
+      APP_ERROR_CHECK( bootloader_dfu_start(_ota_dfu, 3000, true) );
```
hathach, line 255: please add a  line of  descriptive comment such as

```
// DFU timeout in 3 seconds if USB is not enumerated
```

mmoskal:
Thank for comments, updated!

PR #145, opened 2020-07-14
add-ikigaisense_vita-nrf52840

IkigaiSense:
Create the UF2 bootloader for Vita board.

### DIFF
--- a/src/boards/ikigaisense_vita/board.h
+++ b/src/boards/ikigaisense_vita/board.h
@@ -1,0 +1,69 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2018 Ha Thach for Adafruit Industries
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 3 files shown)


src/boards/ikigaisense_vita/board.h
```diff
@@ -0,0 +1,69 @@
...
+//--------------------------------------------------------------------+
+#define BLEDIS_MANUFACTURER   "IkigaiSense Technologies"
+#define BLEDIS_MODEL          "Vita nRF52840"
+
+//--------------------------------------------------------------------+
+// USB
+//--------------------------------------------------------------------+
+#define USB_DESC_VID           0x239A
+#define USB_DESC_UF2_PID       0x0093
+#define USB_DESC_CDC_ONLY_PID  0x0094
```
hathach, line 61: please make this the same to USB_DESC_UF2_PID as well i.e both 0x0093, we are moving away from dual PID for bootloader. It will eventually removed later on.

PR #149, opened 2020-07-17
Fixing Windows not playing nicely with mkdir.

nitz:
I really don't know why Windows is such a pain sometimes:
 * The `-p` doesn't get handled as a switch, and instead actually creates a directory named "-p"...
 * The forward slash in the path being created causes cmd.exe to be sad instead of interpreting it as a path. Quotes around the path makes it happy.

One day things will be sane on Windows.

Today is not that day. 😂

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,8 +47,14 @@ GDB     = $(CROSS_COMPILE)gdb
 NRFUTIL = adafruit-nrfutil
 NRFJPROG = nrfjprog
 
-MK = mkdir -p
-RM = rm -rf
+# Set make directory command, Windows tries to create a directory named "-p" if that flag is there.
+ifneq ($(OS), Windows_NT)
+MK := mkdir -p
+else
+MK := mkdir
+endif
+
+RM := rm -rf
 
 # auto-detect BMP on macOS, otherwise have to specify
 BMP_PORT ?= $(shell ls -1 /dev/cu.usbmodem????????1 | head -1)
@@ -324,7 +330,7 @@ print-%:
 
 # Create build directories
 $(BUILD):
-	@$(MK) $@
+	@$(MK) "$@"
 
 clean:
 	@$(RM) $(BUILD)
--- a/Makefile
+++ b/Makefile
@@ -49,12 +49,12 @@ NRFJPROG = nrfjprog
 
 # Set make directory command, Windows tries to create a directory named "-p" if that flag is there.
 ifneq ($(OS), Windows_NT)
-MK = mkdir -p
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -47,8 +47,14 @@ GDB     = $(CROSS_COMPILE)gdb
 NRFUTIL = adafruit-nrfutil
 NRFJPROG = nrfjprog
 
-MK = mkdir -p
-RM = rm -rf
+# Set make directory command, Windows tries to create a directory named "-p" if that flag is there.
+ifneq ($(OS), Windows_NT)
+MK := mkdir -p
```
hathach, line 52: look good, though why you use `:=` instead of `=`, if there is not much of difference, we should keep it as `=`

nitz: I agree completely. It must have been from when I was tinkering to figure out exactly what was making it upset and I just forgot to change it back!

Makefile
```diff
@@ -47,8 +47,14 @@ GDB     = $(CROSS_COMPILE)gdb
...
 
+# Set make directory command, Windows tries to create a directory named "-p" if that flag is there.
+ifneq ($(OS), Windows_NT)
 MK = mkdir -p
-RM = rm -rf
+else
+MK = mkdir
+endif
+
+RM := rm -rf
```
hathach, line 57: RM `:=` should be reverted as well :)

Makefile
```diff
@@ -47,8 +47,14 @@ GDB     = $(CROSS_COMPILE)gdb
 NRFUTIL = adafruit-nrfutil
 NRFJPROG = nrfjprog
 
+# Set make directory command, Windows tries to create a directory named "-p" if that flag is there.
+ifneq ($(OS), Windows_NT)
 MK = mkdir -p
```
hathach, line 52: if possible, could you please put 2 spaces indent to the MK in the `ifneq`, it makes thing a bit easier to read. This is totally optional

Makefile
nitz: Doh! Willfix.

Makefile
nitz: Done&done! I actually had intended to do that, but used a tab, and was met with a "recipe started before target" and forgot to go back and explore why it wouldn't let me do that.

But I promise I won't get on my "white space shouldn't define structure in languages" soapbox today 🙂

hathach: yeah, tab will trigger makefile to execute recipe rule, space is harmless. It is really frustrating :)

nitz: Well there you have it, I learn something new every day!

Issue #146, opened 2020-07-14
Add unofficial support for ANT and Dual stack softdevices


huddy987:
Currently, if we want to build the bootloader with ANT or dual stack softdevices, we have to resort to hacking our own solution as @orrmany has done. From my understanding the current issue is that the ANT/Dual softdevices aren't allowed to be distributed, and so pre-patched bootloaders with the ANT/Dual softdevices can't be distributed either. My proposition is that we allow new targets in the makefiles to build bootloaders with these softdevices, but leave it up to the user to download the appropriate softdevice.

My idea is to change the makefile from:
make BOARD=XXXXX <Target>
to
make BOARD=XXXXX STACK=XXXXX <Target>
where stack is either "BLE" "ANT" or "BOTH"

When BLE is selected, we can handle building the same way we currently do.

For ANT, it's actually pretty simple. The S212 7.0.1 softdevice supports nRF52832, nRF52810, nRF52833, and nRF52840. So we can use the S212 softdevice as a cover-all solution.

For dualstack, we would need to use S332 for nRF52832, and S340 for nRF52840. We could use 6.1.1, or update to 7.0.1 (doesn't really matter to me).

In any case, the main change would be to create linker scripts for each of these bootloaders, and compile switch in out BLE and ANT code for the single stack softdevices. If a user tries to build without having the ANT/dual stack softdevices, we can just prompt the user to download the stack from thisisant.com.

Of course, this would have to remain as unofficially supported because we can't actually distribute the proposed "new" bootloaders. Users of the "new" bootloaders would be on their own to interface with the softdevices ANT APIs in their apps. All this change would mean is that using ANT with adafruit's arduino library is a bit less tedious.

Thoughts? I took a look at this last night, and thanks to the decoupling of the SD and bootloader, it shouldn't be too hard to do this. If this seems like a reasonable idea I can do the PR.

hathach:
SoftDevice is decoupled and not considered as part of bootloader anymore. You could just update it using uf2 file of any Sxxx you like. Feature is merged, just not released yet. ref https://github.com/adafruit/Adafruit_nRF52_Bootloader/pull/128

huddy987:
There is some BLE code in the bootloader that won't work with the single ANT stack softdevice as well as sd_softdevice_init requires the ANT license parameter...it would be nice to at least have compile switch to compile in/out the required ANT changes

hathach:
[...]

PR #151, opened 2020-07-19
Add dualstack softdevice support

huddy987:
Due to licensing, add a compile switch for creating dualstack softdevice builds. Anyone who wishes to build the dualstack-supported bootloader MUST obtain their own copy of the softdevice from thisisant.com. For evaluation purposes this is free, but anyone who wishes to use the dualstack softdevice commerically must purchase an ANT license.

These builds should be cross-compatible within dualstack softdevices, as long as they are compiled against the dualstack headers (i.e. if the bootloader is built against the s340 headers, the user SHOULD be able to load any s3xx series softdevice and maintain all bootloader functionality).

Single ANT stack support is not added. I was able to disable OTA DFU, but was not able to properly re-enable seiral DFU. This is something I can look at in the future if anyone files an issue requesting it.

Verifed with S140 and S340 v6.1.1 that I am able to:
1. Jump to app
2. Enter serial DFU mode
3. Enter OTA DFU mode and connect via nRF connect

One note on this: If it wasn't for the licensing issue we could have made this a runtime check, but defining ANT_LICENSE_KEY is not allowed by us (the user must download the softdevice and define that themselves in the correct API file).


huddy987:
Build is failing because of https://github.com/adafruit/Adafruit_nRF52_Bootloader/pull/149

hathach:
as said in #146 , Softdevice is not part of bootloader anymore. There is no need to build and merged with it, you just flashed it with jlink or uf2. Unless there is anything else I am not aware of, this PR is not likely to got merged.

huddy987:
@hathach sd_softdevice_enable will fault on S3xx builds if you do not include the ANT_LICENSE_KEY argument. It does not matter if it is or is not a part of the bootloader, you need to build against the correct softdevice headers to get proper bootloader functionality on S3xx softdevices.

Without this change OTA DFU will not be functional with dualstack softdevices.

src/main.c
```diff
@@ -328,7 +328,13 @@ static uint32_t softdev_init(bool init_softdevice)
       .accuracy     = NRF_CLOCK_LF_ACCURACY_250_PPM
   };
 
+#ifdef BLE
   APP_ERROR_CHECK( sd_softdevice_enable(&clock_cfg, app_error_fault_handler) );
+#elif defined(BOTH)
+  APP_ERROR_CHECK( sd_softdevice_enable(&clock_cfg, app_error_fault_handler, ANT_LICENSE_KEY ) );
```
hathach, line 334: I didn't see where `ANT_LICENSE_KEY` is defined anywhere. Are you putting it in the `board.h` ?

hathach (changes requested):
I see, there is a legit reason to add make option, though I preferred to intput the actual `SD=`  for actual softdevice stack than the `PROTOCOL`. That way when Nordic introduce new stack name for new chip (e.g optimized version for smaller chip like 810 we don't have to maintain the makefile). Here is what i think  the PR should implement 
- rename current `SD_NAME` to simply `SD`
- if `SD` is not input, default as it is i.e s132 for 832, s140 for 840
- `SD` is passed to make will change the SD flag, name, linker etc as the same convention as current script

src/main.c
huddy987: ANT_LICENSE_KEY is defined in nrf_sdm.h for the dualstack softdevices. It is commented out by default as a notice that it is only to be used for evaluation purposes. Therefore, I don't think it would be legal for myself to define this.

When a user downloads the correct softdevice headers and compiles against them, this will be defined.

huddy987:
Agreed. I will update the PR to allow users to enter an SD name.

src/main.c
hathach: thanks, this is not too obvious to me

huddy987:
With this change it should be pretty easy to add new softdevices. We just need to add a new "else ifeq" for the new softdevice then add another "else ifeq" for the SoCs that support the new softdevice.

README.md
```diff
@@ -109,25 +109,25 @@ Prerequisites
 To build:
 
 ```
-make BOARD=feather_nrf52840_express all
+make BOARD=feather_nrf52840_express SD=s140 all
```
hathach, line 112: please remove those `SD=s140`, you can add an additional build command to demonstrate using your ANT stack e.g

```
By default, the BLE stack S132/S140 is used, to build with a different softdevice stack, 

make SD=.... 
```

src/main.c
```diff
@@ -328,7 +328,13 @@ static uint32_t softdev_init(bool init_softdevice)
       .accuracy     = NRF_CLOCK_LF_ACCURACY_250_PPM
   };
 
+#ifdef BLE
```
hathach, line 331: I don't like to have an extra macro for this, please use the ANT_LICENSE_KEY for this as well. These macro should be removed from the makefile as well.

```
ifndef ANT_LICENSE_KEY
```

Makefile
```diff
@@ -77,21 +77,71 @@ BUILD = _build/build-$(BOARD)
...
+  $(warning Ensure you download a dualstack softdevice from thisisant.com)
+  $(warning Make sure to compile against the dualstack softdevice headers)
+  CFLAGS += -DS332 -DBOTH
+  DFU_DEV_REV = 0xADAF
+  ifeq ($(MCU_SUB_VARIANT),nrf52)
+    CFLAGS += -DNRF52832_XXAA
+  else
+    $(error Sub Variant $(MCU_SUB_VARIANT) is invalid for softdevice $(SD))
+  endif
+else ifeq ($SD,)
```
hathach, line 125: Too many duplication here, this should be checked before the whole if, and SD should get the default only to S132/S140 only. Then the ifeq() sequence take care of all the CFLAGS.

README.md
```diff
@@ -139,14 +139,34 @@ Supported boards are: feather_nrf52840_express feather_nrf52840_express pca10056
...
+### Building with an ANT softdevice
+
+Currently, the bootloader can be built against the dualstack softdevice headers for
+use of both ANT and BLE simultaneously. To do this:
+
+1. Download either s332 or s340 from thisisant.com. Note that this softdevice is
+   freely available for evaluation use only. Garmin Canada must be contacted to obtain
+   ANT licenses for commercial use.
+2. Place the contents of the softdevice package in the appropriate lib/softdevice folder.
+3. Rename the API folder to <SD name>_nrf52_6.1.1_API.
```
hathach, line 153: can you tell me what is the default name of the s332 and s340 name, I would prefer to have the makefile to use their default name from the extracted instead.

hathach (changes requested):
It is looking good, though we still need a bit of changes for handling the default SD name and the removal of BLE macro.

hathach:
should be implemented by https://github.com/adafruit/Adafruit_nRF52_Bootloader/pull/359

PR #152, opened 2020-07-26
Enable board-specific UF2 family IDs

### LINKED ISSUE #130: Accept uf2 generated with family ID = VID/PID as board specific
**Is your feature request related to a problem? Please describe.**

together with #129, accepting additional family as board specific (VID/PID) allow us to bundle multiple firmware of different boards into a single big uf2 (e.g a circuitpyhon uf2 file for all nRF52 boards ). Following is the original discussion

>FamilyID is meant as both board-id and MCU-id - for example, binaries generated from MakeCode Arcade would run on all NRF54840 boards provided they have the right bootloader settings (that define pinout, screen type, etc.). OTOH some other binaries may have pin configuration baked in, so they will only work on say 840 Feather. So it actually makes perfect sense for the bootloader to accept MCU-wide familyID and a specific boardID.
>
>Generally, this can thought of as a sub-divsion tree of devices and bootloader accepts everything on one path from root to leaf. In this case I think we have 3 levels (all UF2 file - familyID==0, NRF52840, specific board), but it could be more if needed (eg different revisions of a specific board).

_Originally posted by @mmoskal in https://github.com/adafruit/Adafruit_nRF52_Bootloader/pull/128_

Bootloader accept family ID = `ada52840` as generic nrf binary and also family id = vid+pid as board specific one.

**Describe alternatives you've considered**
Application can also use pinconfig in CF2 of bootloader to write generic binary for all boards.

henrygab:
**_NOTE: REQUIRES TESTING_**  (will edit if changes)

Fix #130.

For boards that do not define both USB VID and PID, no change in behavior.  Otherwise...

* Accept UF2 blocks that have the USB VID/PID as the family ID, **_in addition to_** the "legacy" family id of `0xada52840`.

* When exposing the current user application, exposes it using the USB VID/PID as the family ID.

### DIFF
--- a/src/usb/uf2/ghostfat.c
+++ b/src/usb/uf2/ghostfat.c
@@ -363,12 +363,7 @@ void read_block(uint32_t block_no, uint8_t *data) {
                 bl->targetAddr = addr;
                 bl->payloadSize = UF2_FIRMWARE_BYTES_PER_SECTOR;
                 bl->flags = UF2_FLAG_FAMILYID;
-
-#if defined(CFG_UF2_BOARD_APP_ID) && CFG_UF2_BOARD_APP_ID
                 bl->familyID = CFG_UF2_BOARD_APP_ID;
-#else
-                bl->familyID = CFG_UF2_FAMILY_APP_ID;
-#endif
                 memcpy(bl->data, (void *)addr, bl->payloadSize);
             }
         }
@@ -395,12 +390,8 @@ int write_block (uint32_t block_no, uint8_t *data, WriteState *state)
 
   switch ( bl->familyID )
   {
-
-#if defined(CFG_UF2_BOARD_APP_ID) && CFG_UF2_BOARD_APP_ID
     case CFG_UF2_BOARD_APP_ID:  // board-specific app ... may not be usable on other nrf52 boards
-#endif
-
-    case CFG_UF2_FAMILY_APP_ID: // legacy, or where app uses bootloader configuration to discover pins
+    case CFG_UF2_FAMILY_APP_ID: // family-specific app ... Legacy apps, or where the app uses bootloader configuration to discover pins
       /* Upgrading Application
        *
        * SoftDevice is considered as part of application and can be (or not) included in uf2.
--- a/src/usb/uf2/ghostfat.c
+++ b/src/usb/uf2/ghostfat.c
@@ -363,12 +363,7 @@ void read_block(uint32_t block_no, uint8_t *data) {
                 bl->targetAddr = addr;
                 bl->payloadSize = UF2_FIRMWARE_BYTES_PER_SECTOR;
                 bl->flags = UF2_FLAG_FAMILYID;
-
-#if defined(CFG_UF2_BOARD_APP_ID) && CFG_UF2_BOARD_APP_ID
                 bl->familyID = CFG_UF2_BOARD_APP_ID;
-#else
-                bl->familyID = CFG_UF2_FAMILY_APP_ID;
-#endif
                 memcpy(bl->data, (void *)addr, bl->payloadSize);
             }
         }
--- a/src/usb/uf2/ghostfat.c
+++ b/src/usb/uf2/ghostfat.c
@@ -363,7 +363,13 @@ void read_block(uint32_t block_no, uint8_t *data) {
                 bl->targetAddr = addr;
                 bl->payloadSize = UF2_FIRMWARE_BYTES_PER_SECTOR;
                 bl->flags = UF2_FLAG_FAMILYID;
-                bl->familyID = CFG_UF2_FAMILY_APP_ID;
+
+                bool useBoardId = false; // BUGBUG -- how to detect which familyID to use?
+                if (useBoardId) {
+                  bl->familyID = CFG_UF2_BOARD_APP_ID;
+                } else {
+                  bl->familyID = CFG_UF2_FAMILY_APP_ID;
+                }
                 memcpy(bl->data, (void *)addr, bl->payloadSize);
             }
         }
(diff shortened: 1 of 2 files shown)


src/usb/uf2/uf2cfg.h
```diff
@@ -1,9 +1,15 @@
 #include "boards.h"
 #include "dfu_types.h"
 
-// Family ID for updating Application
+// Legacy Family ID for updating Application
 #define CFG_UF2_FAMILY_APP_ID     0xADA52840
 
+// Family ID for board-specific Application
+#if defined(USB_DESC_VID) && defined(USB_DESC_UF2_PID) && USB_DESC_VID && USB_DESC_UF2_PID
```
hathach, line 8: let's make the CFG_UF2_BOARD_APP_ID always defined. This repo doesn't accept nrf52840/833 that doesn't have VID/PID.

src/usb/uf2/ghostfat.c
```diff
@@ -390,7 +395,12 @@ int write_block (uint32_t block_no, uint8_t *data, WriteState *state)
 
   switch ( bl->familyID )
   {
-    case CFG_UF2_FAMILY_APP_ID:
+
+#if defined(CFG_UF2_BOARD_APP_ID) && CFG_UF2_BOARD_APP_ID
+    case CFG_UF2_BOARD_APP_ID:  // board-specific app ... may not be usable on other nrf52 boards
```
hathach, line 400: should always exists.

src/usb/uf2/uf2cfg.h
```diff
@@ -1,9 +1,15 @@
 #include "boards.h"
 #include "dfu_types.h"
 
-// Family ID for updating Application
+// Legacy Family ID for updating Application
```
hathach, line 4: it is not legacy, it is generic one, uf2 with 
- vid/pid can be upgraded
- family =  generic ada52840 should also be accepted as well.

src/usb/uf2/ghostfat.c
```diff
@@ -363,7 +363,12 @@ void read_block(uint32_t block_no, uint8_t *data) {
                 bl->targetAddr = addr;
                 bl->payloadSize = UF2_FIRMWARE_BYTES_PER_SECTOR;
                 bl->flags = UF2_FLAG_FAMILYID;
+
+#if defined(CFG_UF2_BOARD_APP_ID) && CFG_UF2_BOARD_APP_ID
```
hathach, line 367: yeah, should always return the family = BOARD_ID here.

src/usb/uf2/uf2cfg.h
henrygab: Ok, fixed with commit aa8cd83.

src/usb/uf2/ghostfat.c
henrygab: Ok, fixed with commit aa8cd83.

src/usb/uf2/uf2cfg.h
henrygab: Yes, the code allows both.

src/usb/uf2/ghostfat.c
henrygab: Ok, fixed with commit aa8cd83.

henrygab:
> Thanks @henrygab, welcome back, hopefully you have resolved your real life issues.

Thanks. 😃 

> Indeed the PR is simpler that I thought. [...] I think we can just be bold and define CFG_UF2_BOARD_APP_ID as always existed. The generic FAMILY ID already ensure the backward compatible, and every board with USB required VID/PID defined.

Well, I tried to presume VID/PID are defined with commit aa8cd83.  However, it appears that the feather_nrf52832 board does not define these.  So, I guess I should revert to the prior version, which allows this to not be defined.  😆   

Also, I cannot test this except on feather nrf52840 express, so I will defer to you for testing.

P.S. -- EPS32s2 looks interesting.

src/usb/uf2/uf2cfg.h
hathach: I mean you need to remove the `legacy` in the comment

hathach:
> Well, I tried to presume VID/PID are defined with commit [aa8cd83](https://github.com/adafruit/Adafruit_nRF52_Bootloader/commit/aa8cd83961534464d4bcfec23a6c2c4671869cf5). However, it appears that the feather_nrf52832 board does not define these. So, I guess I should revert to the prior version, which allows this to not be defined. 
832 doesn't have USB, this code won't be included to the CC progress. Please skip those ifndef

src/usb/uf2/uf2cfg.h
hathach: you could keep ifndef here to make CI happy for now. Please remove other in ghostfat.c, that file is not compiled with 832

henrygab:
OK, updated to remove the `#ifdef` from ghostfat.c, and left them in for uf2cfg.h due to build errors for '832.  Ready for testing?

src/usb/uf2/ghostfat.c
```diff
@@ -390,7 +390,12 @@ int write_block (uint32_t block_no, uint8_t *data, WriteState *state)
 
   switch ( bl->familyID )
   {
-    case CFG_UF2_FAMILY_APP_ID:
+
+#if defined(CFG_UF2_BOARD_APP_ID) && CFG_UF2_BOARD_APP_ID
```
hathach, line 394: please remove this #if as well

henrygab: Well, I **_could_** remove that `#if`/`#endif` pair.
However, I don't recommend doing so, as it will break the build for the `832 board, which doesn't define that symbol.

henrygab:
@hathach 
Given all comments addressed, I've updated the title to remove [WIP] tag.
However, I'm leaving the PR as draft because testing is still required, and it will stop accidental merge.

src/usb/uf2/ghostfat.c
hathach: no it wont' this file won't be compiled with 832.The previous issue exist because the define in the header file, which actually shouldn't be included by 832. Though I will fix that later on as mentioned above.

henrygab: _mea culpa!_  you're right ... I'm sorry.  Pushed the change.

hathach:
> @hathach
> Given all comments addressed, I've updated the title to remove [WIP] tag.
> However, I'm leaving the PR as draft because testing is still required, and it will stop accidental merge.

Thanks @henrygab, I will do actual hardware testing next week

Issue #150, opened 2020-07-17
[Bug] Unable to use NFCT


nitz:
**Describe the bug**
When using the UF2 bootloader, I'm unable to use the NFCT peripheral, despite the API calls looking okay.

**Set up (please complete the following information)**
- Using the [nRF52840 M.2 Dev Kit](https://github.com/makerdiary/nrf52840-m2-devkit) from makerdiary. For the board configuration, I'm basically using their fork's version [here](https://github.com/makerdiary/uf2-bootloader/tree/m2/src/boards/nrf52840_m2) but updated to have a pinconfig.c like the other boards. I've attached my current copy here [boards-nrf52840_m2.zip](https://github.com/adafruit/Adafruit_nRF52_Bootloader/files/4935804/boards-nrf52840_m2.zip). (Note the -U CFLAG in the mk, more on that in a bit.)
- Win10 Pro/Workstations 20H2 (19042.388)
- Not much serial log to show, as everything is pretty much generated from my output, but here's what it looks like anyways, once the application is running:
```
<debug> app: =====================================
<info> app: Application starting.
<debug> nfctagemu: Initializing Type 2 Tag NFCT via nrfx. (0x047D)
<info> NFCT: Initialized
<debug> nfctagemu: Payload set @0x200000EC
<info> nfctagemu: Tag enabled.
<info> NFCT: Start
<debug> app: Application waiting for events!
(the NFCT events never come...)
```

**To Reproduce**
1. Set up a board that you plan to use the NFCT on
2. Comment out the `CFLAGS += -DCONFIG_NFCT_PINS_AS_GPIOS` line from the main `Makefile`, so the system init doesn't try to skip the NFCT antenna pins. (I tried using `CFLAGS += -UCONFIG_NFCT_PINS_AS_GPIOS` in the `board.mk` file, but since it's evaluated before the line that *adds* the define, it never ends up being useful. Maybe we could move the CFLAGS defs above the included *.mk files?)
3. Flash the bootloader.
4. Use the bootloader to install an application with which to use the NFCT, using a phone reader software or other, observe it not working.
5. Using a non UF2 flash/bootloader, flash that same application. It works!

**Expected behavior**
I should be able to use the NFCT. I might just be missing something that I've not got set up.

**Screenshots**
N/A

**Additional context**
[...]

PR #155, opened 2020-08-04
Add MakerDiary nRF52840 M.2 Module board.

nitz:
Currently, board.mk has CFLAGS additions attempting to undefine `CONFIG_NFCT_PINS_AS_GPIOS` and `SOFTDEVICE_PRESENT`. Those flags, when used on this board prevent the use of the NFCT, as per #150.

As the PR currently stands, those CFLAGS additions don't work to correctly undefine the symbols, as the board.mk is included ahead of the default CFLAGS being set.

Perhaps we could use another flag similar to how `MCU_SUB_VARIANT` is handled, something like `USES_NFCT`, and `NO_SOFTDEVICE` to prevent those definitions, allowing the individual boards to skip those at will?


nitz:
Something like that, that I pushed in 75c8999?

I'm not thrilled with those variable names, but they are at least somewhat descriptive. 🙃

Makefile
```diff
@@ -257,8 +257,17 @@ CFLAGS += -Wno-cast-function-type
...
-CFLAGS += -DSOFTDEVICE_PRESENT
+
+# Skip defining CONFIG_NFCT_PINS_AS_GPIOS if the device uses the NFCT.
+ifneq ($(USES_NFCT),yes)
+  CFLAGS += -DCONFIG_NFCT_PINS_AS_GPIOS
+endif
+
+# Skip defining SOFTDEVICE_PRESENT if the device doesn't use a soft device.
+ifneq ($(NO_SOFT_DEVICE),yes)
+  CFLAGS += -DSOFTDEVICE_PRESENT
```
hathach, line 268: To completely remove the SD, we should also exclude all the sd headers file as well to prevent hardfault in run time. Since the current code try its best to detect actual SD existed rather than relying on this macro https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/master/lib/sdk11/components/libraries/bootloader_dfu/dfu_types.h#L38

This can help to avoid calling sd API if not actually flashed, but a mistake can accidentally invoke an API, therefore a full header excluding is needed.

It is best to use the current `SD_NAME` variable to do this, by default it is as follow, 
https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/master/Makefile#L91

you should change the `SD_NAME = s140` to `SD_NAME ?= s140`. And having board.mk to define it to `SD=` (null), then modify the include path and macro along with it accordingly.

Makefile
```diff
@@ -257,8 +257,17 @@ CFLAGS += -Wno-cast-function-type
 # Defined Symbol (MACROS)
 CFLAGS += -D__HEAP_SIZE=0
 CFLAGS += -DCONFIG_GPIO_AS_PINRESET
-CFLAGS += -DCONFIG_NFCT_PINS_AS_GPIOS
-CFLAGS += -DSOFTDEVICE_PRESENT
+
+# Skip defining CONFIG_NFCT_PINS_AS_GPIOS if the device uses the NFCT.
+ifneq ($(USES_NFCT),yes)
```
hathach, line 262: would you mind changing it to just `USE_NFCT`, I found it easier to read with all cap.

src/boards/nrf52840_m2/board.h
```diff
@@ -0,0 +1,70 @@
...
+//--------------------------------------------------------------------+
+// BLE OTA
+//--------------------------------------------------------------------+
+#define BLEDIS_MANUFACTURER   "cmd.wtf"
+#define BLEDIS_MODEL          "PowerOverwhelmiing"
+
+//--------------------------------------------------------------------+
+// USB
+//--------------------------------------------------------------------+
+#define USB_DESC_VID            0x1915
```
hathach, line 59: Do you have the permission from Nordic to use this VID/PID ?

nitz: I'm not sure, but I'm guessing no. I just duplicated the data from Makerdiary's fork, in which they used this VID/PID. I see in some other boards it looks like you're sharing the VID/PID of the feather. Is that ideal for devices that don't have their own unique ones?

Makefile
nitz: This is a much better solution. Will push a commit in a bit for this, but it will take me a short bit as there's a few places this touches!

Makefile
nitz: Definitely. I like the way that reads better too.

Makefile
nitz: I actually see that makerdiary did some work to disable softdevice in their branch too, so I'm looking at their commits now to see the sorts of changes I'm needing to make. I am going to remove this one from this PR for the time being, and open a separate PR for that, as it seems it should be a change on it's own unrelated to this device!

nitz:
Okay, this is good to go as soon as we can resolve the VID/PID bit. I'm unsure what the standard is for boards that don't have "valid" (is that the right word?) IDs.

src/boards/nrf52840_m2/board.h
nitz: I realize now I commited these changes with string I was using for my device I'm tinkering with and not the ones from MakerDiary. Somehow I managed to get my branches confused! Changed now, in 8ab7a41.

Still not sure what to do about the VID/PID though -- I've looked around and can't see any public listings Nordic keeps of PID's they've handed out for their VID.

nitz: And as I continue to ramble on....

I've found these posts [here](https://devzone.nordicsemi.com/f/nordic-q-a/33851/nrf52840---usb-vendor-id-and-usb-product-id) and more specifically [this one](https://devzone.nordicsemi.com/f/nordic-q-a/30440/nrf24lu1-pid-vid/120663#120663), which makes it seem like Nordic allows open use of their VID, as long as you're shipping on a Nordic chip. They just don't track PIDs so they may collide, but that seems like a small issue for us!

hathach: That will cause conflict in the future, it is still better to share a generic VID/PID bootloader than conflicting with other application mode. Since most bootloader behaves (and have the same usb interfaces). Therefore you have 2 options
1. Try to get circuitpython running on your board, and submit PR to its repo. You are eligible to get a free VID/PID from Adafruit. There is lots of PR/issues regarding this in circuipython. You can take a look there.
2. Wait for me to make some changes (with some internal discussion with Adafruit team) to decide which VID/PID can be used as generic one. 

Except for some early board sharing the VID/PID. Most of the boards that has unique here are donated one from Adafruit. It is relatively easy proess, you can try it out.

nitz: So as I started looking around, it turns out it actually already is mainlined in circuitpython as of about a month ago it looks like: https://github.com/adafruit/circuitpython/blob/main/ports/nrf/boards/makerdiary_nrf52840_m2_devkit/mpconfigboard.mk

They're using a PID/VID that was granted to them by Seeed, it seems (0x2886/0xF001)!

So I looked at a couple of the adafruit boards between circuitpython and here. It seems they all use the same PID for the bootloader vs circuitpython's, except circuitpython's has the MSB 1 and the bootloader's has the MSB 0?

Should I see about asking adafruit for a bootloader-specific PID on the adafruit VID? Or should I instead try to get in touch with @makerdiary to see if perhaps Seeed had given them some other PIDs on the Seeed VID to use for bootloaders? I will tweet them with this comment just in case anyways!

xiongyihui:
@nitz We got a range of PIDs from Seeed Studio. We can use `0x2886:0xf00f` for this M2 module.

nitz:
> @nitz We got a range of PIDs from Seeed Studio. We can use `0x2886:0xf00f` for this M2 module.

Awesome, thank you, @xiongyihui! I plan to add your Pitaya Go board in the future, too. I saw you're still using the Nordic VID/PID in [circuitpython](https://github.com/adafruit/circuitpython/blob/main/ports/nrf/boards/pitaya_go/mpconfigboard.mk). Any chance you had a pair of PIDs you wanted to give to that board as well?

@hathach This one should be good to go!

xiongyihui:
@nitz  For the bootloader of Pitaya Go, we can use `0x2886:0xf00e`. Nordic allows small vendors to use its VID and PIDs . The issue is that one PID may be used by two or more products, which may cause windows driver problem.

nitz:
> @nitz For the bootloader of Pitaya Go, we can use `0x2886:0xf00e`. Nordic allows small vendors to use its VID and PIDs . The issue is that one PID may be used by two or more products, which may cause windows driver problem.

Excellent, will do! Thank you @xiongyihui

Issue #115, opened 2020-03-10
[Bug] Upgrading bootloader to 0.3.1 using jlink ended up in bootloader loop


akamensky:
**Describe the bug**
I decided to upgrade bootloader using my jlink from 0.3.0 to 0.3.1. After writing, I still "can" upload arduino sketches and other compiled binaries, however it is just stuck in bootloader (green led light up) and mass storage volume shows up repeatedly (every couple of minutes) named "FTHR840BOOT" with usual UF2 volume contents. Uploading any UF2 files to that volume results in no effect (no unmount and flash happens, it just stays there as mass volume).

Trying to revert to 0.3.0 has no effect, it still stuck in bootloader loop and shows "new" volume name.

Trying to do UICR erase using nrfjprog or recovery still leads to the same outcome after flashing bootloader.

**Set up (please complete the following information)**
  - feather_nrf52840_express
  - macos
  - ...

hathach:
try to upload and uf2 or an sketch via Serial once to settle the DFU setting. It should work afterward with jlink upload

akamensky:
I've recovered board somewhat, but it keeps going to bootloader after running valid sketch for some time (at random it looks).

hathach:
it is not a new bug, it is CRC checksum of bootloader, you should either flash the CRC16 of the application or disable the checking by writing 0x0001 (if I remembered correctly) to bootloader crc checking address (forget the exact address). Since you just erase and flash with jlink, simplest way is perfrom an DFU via serial or uf2, it iwll write the disable magice to crc checking.

henrygab:
I had this same problem ... and I can confirm that installing a UF2 (I used [Circuit Python](https://circuitpython.org/board/pca10056/) as it was pre-generated) solved this bootloop for me just now.  What a relief!  I thought I had somehow damaged my DK, but it now boots sketches correctly again.  YAY!

PR #156, opened 2020-08-04
Add definable flag to skip app CRC check.

nitz:
If a board.mk defines a symbol named `DISABLE_APPLICATION_CRC_CHECK` (such as via `CFLAGS += -DDISABLE_APPLICATION_CRC_CHECK`), the code that would otherwise check for the application's CRC validity will not be compiled, and success will be assumed.

This would be an optional feature for users like in #115 that don't have CRC calculation (and setting) as part of their flash process.

### DIFF
--- a/lib/sdk11/components/libraries/bootloader_dfu/bootloader.c
+++ b/lib/sdk11/components/libraries/bootloader_dfu/bootloader.c
@@ -160,7 +160,7 @@ bool bootloader_app_is_valid(void)
   {
     return false;
   }
-
+#if !defined(DISABLE_APPLICATION_CRC_CHECK)
   // The application in CODE region 1 is flagged as valid during update.
   if ( p_bootloader_settings->bank_0 == BANK_VALID_APP )
   {
@@ -176,6 +176,10 @@ bool bootloader_app_is_valid(void)
 
     success = (image_crc == p_bootloader_settings->bank_0_crc);
   }
+#else
+  // DISABLE_APPLICATION_CRC_CHECK was defined, assume CRC check passed.
+  success = true;
+#endif
 
   return success;
 }


nitz:
Okay, I'm stumped as to why that build would fail. It only failed on the one `feather_nrf52832` board, and succeeded on the other 23? And the error had nothing to do with the changed file. `build_all.py` builds all 24 just fine locally here...

hathach:
> Okay, I'm stumped as to why that build would fail. It only failed on the one `feather_nrf52832` board, and succeeded on the other 23? And the error had nothing to do with the changed file. `build_all.py` builds all 24 just fine locally here...

action ci can fail randomly under heavy load, reset the run would fix that. I will review the PR later on.

nitz:
Ah, this shows my poor understanding of some of the tooling I'm using here! 🙂

I guess this brings the question then: Is this the sort of thing you'd want to be supported via the bootloader makefile then? Seeing how circuitpython does it by directly poking the 1 into the settings address, do you feel it should be an option similar to that in the bootloader? Or should it just be up to the application/user?

Also not directly related: I'm using pyocd to flash, so I'm building then flashing in two separate steps. Would a 'FLASHER' option like circuitpython uses to choose between $(NRFJPROG) and pyocd be acceptable?

hathach:
> Ah, this shows my poor understanding of some of the tooling I'm using here! 
> 
> I guess this brings the question then: Is this the sort of thing you'd want to be supported via the bootloader makefile then? Seeing how circuitpython does it by directly poking the 1 into the settings address, do you feel it should be an option similar to that in the bootloader? Or should it just be up to the application/user?

I think we should better to go through the code to disable it. It will take a bit more than your changes to do that, I am just too lazy for now to do so. Having supported via bootlaoder makefile is kind of awkward I think. If you don't want to spend time to proceed with the PR, feel free to open an issue as reminder, I will do it whenever I have time.

> Also not directly related: I'm using pyocd to flash, so I'm building then flashing in two separate steps. Would a 'FLASHER' option like circuitpython uses to choose between $(NRFJPROG) and pyocd be acceptable?

having pyocd as additional FLASHER is welcome, many people use cheap DAPLink such as particle debugger or on-board. I don't see any issues having more flasher except for work hour spent on the PR :smile:

nitz:
>I think we should better to go through the code to disable it. It will take a bit more than your changes to do that, I am just too lazy for now to do so. Having supported via bootlaoder makefile is kind of awkward I think. If you don't want to spend time to proceed with the PR, feel free to open an issue as reminder, I will do it whenever I have time.

Yeah, I think you're right. I'm gonna toss this one on my backburner for now, since just poking 1 in the settings address will work for the time being. I'll open an issue just to keep it on the radar!

>having pyocd as additional FLASHER is welcome, many people use cheap DAPLink such as particle debugger or on-board. I don't see any issues having more flasher except for work hour spent on the PR 😄

Excellent, I think I will work on this, this evening then. I'm finding myself wanting to keep my application and bootloader pretty separate at this point still, so being able to easily flash from the BL project will be nice to do all in one step. And you said it, I'm a big fan of the cheap DAPLink stuff, at least until I put on my big person britches and shell out for a real J-Link. (Cause of course every time I go look at them, I end up coveting the most expensive one 🙃)

PR #158, opened 2020-08-06
Add pyocd flash targets.

nitz:
They are used when make is invoked with FLASHER=pyocd. Still defaults to using NRFUTIL.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,12 +47,11 @@ GDB     = $(CROSS_COMPILE)gdb
 # Flasher utility options
 NRFUTIL = adafruit-nrfutil
 NRFJPROG = nrfjprog
-FLASHER ?=
+FLASHER ?= nrfjprog
 
-# default to nrf
-ifeq ($(FLASHER),)
-  FLASHER = nrf
-else ifeq ($(FLASHER),pyocd)
+# Flasher will default to nrfjprog,
+# Check for pyocd, error on unexpected value.
+ifeq ($(FLASHER),pyocd)
   PYOCD ?= pyocd
 else
   $(error Unsupported flash utility: "$(FLASHER)")
@@ -398,8 +397,8 @@ __check_defined = \
     $(if $(value $1),, \
     $(error Undefined make flag: $1$(if $2, ($2))))
 
-#------------------- Flash with NRFUTIL -------------------
-ifeq ($(FLASHER),nrf)
+#------------------- Flash with NRFJPROG -------------------
+ifeq ($(FLASHER),nrfjprog)
 
 # Flash the compiled
 flash: $(BUILD)/$(OUT_FILE)-nosd.hex
--- a/Makefile
+++ b/Makefile
@@ -398,12 +398,11 @@ __check_defined = \
     $(if $(value $1),, \
     $(error Undefined make flag: $1$(if $2, ($2))))
 
-	
 #------------------- Flash with NRFUTIL -------------------
 ifeq ($(FLASHER),nrf)
 
-  # Flash the compiled
-  flash: $(BUILD)/$(OUT_FILE)-nosd.hex
+# Flash the compiled
+flash: $(BUILD)/$(OUT_FILE)-nosd.hex
 	@echo Flashing: $(notdir $<)
 	$(NRFJPROG) --program $< --sectoranduicrerase -f nrf52 --reset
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -387,32 +398,66 @@ __check_defined = \
...
+	@echo Flashing: $(MBR_HEX)
+	$(PYOCD) flash -t $(MCU_SUB_VARIANT) $(MBR_HEX)
+
+endif
+
+#------------------- Debugging -------------------
+
 gdbflash: $(BUILD)/$(MERGED_FILE).hex
-	@echo Flashing: $<
+ 	@echo Flashing: $<
```
hathach, line 460: this rule must remain to use tab

Makefile
```diff
@@ -387,32 +398,66 @@ __check_defined = \
...
     $(error Undefined make flag: $1$(if $2, ($2))))
 
-# Flash the compiled
-flash: $(BUILD)/$(OUT_FILE)-nosd.hex
+	
+#------------------- Flash with NRFUTIL -------------------
+ifeq ($(FLASHER),nrf)
+
+  # Flash the compiled
+  flash: $(BUILD)/$(OUT_FILE)-nosd.hex
```
hathach, line 406: You should not indent the target name, just leave it as it is

Makefile
```diff
@@ -44,8 +44,19 @@ OBJCOPY = $(CROSS_COMPILE)objcopy
 SIZE    = $(CROSS_COMPILE)size
 GDB     = $(CROSS_COMPILE)gdb
 
+# Flasher utility options
 NRFUTIL = adafruit-nrfutil
 NRFJPROG = nrfjprog
+FLASHER ?=
+
+# default to nrf
+ifeq ($(FLASHER),)
```
hathach, line 53: `FLASHER ?= nrfjprog` is preferable than null

Makefile
```diff
@@ -387,32 +398,66 @@ __check_defined = \
     $(if $(value $1),, \
     $(error Undefined make flag: $1$(if $2, ($2))))
 
-# Flash the compiled
-flash: $(BUILD)/$(OUT_FILE)-nosd.hex
+	
+#------------------- Flash with NRFUTIL -------------------
```
hathach, line 402: Flash with nrfjprog to be precise, nrfutil is for dfu-flash only.

Makefile
```diff
@@ -387,32 +398,66 @@ __check_defined = \
     $(if $(value $1),, \
     $(error Undefined make flag: $1$(if $2, ($2))))
 
-# Flash the compiled
-flash: $(BUILD)/$(OUT_FILE)-nosd.hex
+	
+#------------------- Flash with NRFUTIL -------------------
+ifeq ($(FLASHER),nrf)
```
hathach, line 403: nrfjprog is more appropriate since we are referring to the flash tool/command. nrf is too generic.

Makefile
```diff
@@ -44,8 +44,19 @@ OBJCOPY = $(CROSS_COMPILE)objcopy
...
+# Flasher utility options
 NRFUTIL = adafruit-nrfutil
 NRFJPROG = nrfjprog
+FLASHER ?=
+
+# default to nrf
+ifeq ($(FLASHER),)
+  FLASHER = nrf
+else ifeq ($(FLASHER),pyocd)
+  PYOCD ?= pyocd
```
hathach, line 56: we could simplify the target rule by define FLASH_CMD function that take hex file as input and expand to appropriate flash command.  Much cleaner than defining each target for each flasher. However If you are not familiar with makefile function define, feel free to skip it, I will do it later on.

Makefile
nitz: Heyyy, how did that space get in there! o.o

nitz:
Okay, I think I got them all!

I'm definitely not super familiar with Makefile functions (but enjoy learning!), and ran into issues trying to do conditionals inside them, passing a 'mode' for FLASH_CMD. I think I might have gotten it to work, but it would have been about 45 parenthesis deep. Is the way I've done there with 3 FLASH_* functions like that something along the lines of what you imagined, or did you have a different idea?

hathach:
> Okay, I think I got them all!
> 
> I'm definitely not super familiar with Makefile functions (but enjoy learning!), and ran into issues trying to do conditionals inside them, passing a 'mode' for FLASH_CMD. I think I might have gotten it to work, but it would have been about 45 parenthesis deep. Is the way I've done there with 3 FLASH_* functions like that something along the lines of what you imagined, or did you have a different idea?

No worries, That is OK to have 3 flash functions, we could refactor it later on when there are more FLAHSER, the code is good enough since it is more centric to look at, and more importantly we don't have to duplicate target for each new FLASHER. It is perfect now :)

PR #160, opened 2020-08-17
Resolve OTA failure with bond shared from application

### LINKED ISSUE #104: [Bug]Cannot OTA when paired using Android device
I am using feather_nrf52832 board running the latest bootloader and the BLE UART example from here https://learn.adafruit.com/bluefruit-nrf52-feather-learning-guide?view=all It works fine with Android without paring, but with pairing, the board will reboot. But it will go into the application other than doing the update. I am updating using nRFtoolbox and nRF connect app. The error message from both is (0x85) GATT Error. 

Is this a feature not implemented?

KK1423:
Fixes #104 and allows for OTA DFU update with peers that are already bonded.


hathach (approved):
Somehow I totally missed this PR. Thank you for your fix. This looks good to me.

PR #166, opened 2020-09-04
remove dfu timeout

### LINKED ISSUE #164: Why boot drive timeout?
With this UF2bootloader (not the Atmel one), there is a timeout after several minutes when the boot drive disappears. Currently this seems to be causing a Finder crash in some circumstances. See https://forums.adafruit.com/viewtopic.php?f=65&t=168987.

@hathach Is the timeout deliberate? What was the motivation? Thanks.

hathach:
fix #164

### DIFF
--- a/lib/sdk11/components/libraries/bootloader_dfu/bootloader.c
+++ b/lib/sdk11/components/libraries/bootloader_dfu/bootloader.c
@@ -87,7 +87,8 @@ static void dfu_startup_timer_handler(void * p_context)
   }
 #endif
 
-  // No packets are received within timeout, terminal and DFU mode
+  // nRF52832 forced DFU on startup
+  // No packets are received within timeout, exit DFU mode
   // dfu_startup_packet_received is set by process_dfu_packet() in dfu_transport_serial.c
   if (!dfu_startup_packet_received)
   {
@@ -336,7 +337,9 @@ uint32_t bootloader_dfu_start(bool ota, uint32_t timeout_ms, bool cancel_timeout
       err_code = dfu_transport_ble_update_start();
     }else
     {
-      // timeout_ms > 0 is forced startup DFU
+      // DFU mode with timeout can be
+      // - Forced startup DFU for nRF52832 or
+      // - Makecode single tap reset but no enumerated (battery power)
       if ( timeout_ms )
       {
         dfu_startup_packet_received = false;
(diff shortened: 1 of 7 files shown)


hathach:
hmm @dhalbert ci failed to build, complaining the flash is overflow, I am building just fine locally with the very same version of gcc. I just trigger a rebuild of good-old master, and it failed as well. Very weird issue :thinking: :thinking:

lib/sdk11/components/libraries/bootloader_dfu/bootloader.c
```diff
@@ -336,7 +337,9 @@ uint32_t bootloader_dfu_start(bool ota, uint32_t timeout_ms, bool cancel_timeout
       err_code = dfu_transport_ble_update_start();
     }else
     {
-      // timeout_ms > 0 is forced startup DFU
+      // DFU mode with timeout can be
+      // - Forced startup DFU for nRF52832 or
+      // - Makecode single tap reset but no enumerated (battery power)
       if ( timeout_ms )
       {
```
dhalbert, line 344: It looks like you removed the timeout time-keeping code, but why is there still a `timeout_ms` mention here, and mentions of timeout in the comments? I am just a little confused.

dhalbert (commented):
I tried this on a CLUE that needed a bootloader update anyway. It works fine. It didn't time out after waiting more than 10 minutes, and it successfully loaded CircuitPython. So my comments are just some questions about the remaining mentions of timeouts in the code.

lib/sdk11/components/libraries/bootloader_dfu/bootloader.c
hathach: This is another timeout, which is different than the overall timeout which is just removed. This one is used by nrf52832, 832 always go into DFU mode for 1 second when power on. This is required for Arduino auto reset is hardware to CP2104 DTR. It is later used by Makecode to exit its singletap reset to DFU when on battery power only (no usb enumeration) after 3 seconds.

dhalbert: :heavy_check_mark:

dhalbert:
I merged the PR (maybe I should have waited for you), and the build failed: https://github.com/adafruit/Adafruit_nRF52_Bootloader/runs/1075498601

hathach:
> I merged the PR (maybe I should have waited for you), and the build failed: https://github.com/adafruit/Adafruit_nRF52_Bootloader/runs/1075498601

restart the build should fix this, I think it is caused by parallel compiling with `make -j 4`. I experience something similar when compile parallel with `clean all` in local machine. I am not entirely sure though, it was OK until recently. Don't worry I will look at it later on.

PR #176, opened 2020-12-01
Add optional Makefile.user (eg for BOARD=...)

mmoskal:
One can put `BOARD = feather_nrf52840_express` or similar in `Makefile.user` to set a default


hathach (approved):
look good to me, CI somehow failed to run, just restarted, will be merged after it passes.

PR #177, opened 2020-12-02
Allow for power supply configuration

mmoskal:
In case the MCU is supplied through its own LDO or DCDC, it's useful to configure that from bootloader. Many boards don't have an inductor for the DCDC0, but at least some modules have one for DCDC1, so I made them independent. The power consumption is about 2x lower with DCDC compared to LDO.

### DIFF
--- a/src/boards/boards.c
+++ b/src/boards/boards.c
@@ -82,6 +82,31 @@ void board_init(void)
   neopixel_init();
 #endif
 
+#if ENABLE_DCDC_0 == 1
+  NRF_POWER->DCDCEN0 = 1;
+#endif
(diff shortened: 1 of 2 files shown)


hathach:
Seem like action ci need to change from setup node v1.1.0 to simply v1.1 to pass. I am on 2 days leave. Will merge now and fix the ci later on

mmoskal:
No rush!

hathach (approved):
thank you very much for the PR. To be honest, I haven't tested out the effect of using DCDC on actual hardware. However, since it does not change the behavior with default value and the modification look good enough.

PR #186, opened 2021-01-30
fix against pollution of REGOUT0 reserved bits

lyusupov:
When UICR_REGOUT0_VALUE  macro is set to
```

#define UICR_REGOUT0_VALUE UICR_REGOUT0_VOUT_3V3

```
as suggested by author of https://github.com/adafruit/Adafruit_nRF52_Bootloader/pull/177 - the REGOUT0  new value becomes 0x00000005 instead of 0xFFFFFFFD.

This PR is intended to fix the issue.


hathach (approved):
look great, thank you for your PR.

PR #189, opened 2021-02-14
Add USB connect timeout for app reset to UF2 or Serial via GPREGRET

### LINKED ISSUE #188: [Bug] Stuck in UF2 DFU mode forever when USB not connected
**Describe the bug**
The 0.4.0 release notes state
- Reset into application if there is no USB connection in ~3 seconds.
- Remove DFU idle 300 seconds timeout

The Reset into application without USB connection does not work.

**Set up**
  - nrf52840 Express
  - Bootloader 0.4.0

**To Reproduce**
1. Disconnect USB
2. From application, execute the following to enter UF2 mode (or use some super secret button combo)

```
  NRF_POWER->GPREGRET = 0x57;  // UF2 DFU mode
  sd_nvic_SystemReset();
```

3. Wait for application to restart when no USB connection is established

At this point, the application is stuck in DFU mode forever.  It does not 'time out' after 3 seconds (or 300 seconds)

**Expected behavior**
The device should restart in application mode after 3 seconds if no USB connection.

**Additional context**
My device is embedded inside an enclosure without access to the reset button, tricky double button presses, triple clicks and so forth.  It does have a way though via a user input to initiate UF2 mode to allow upgrades.  Without a timeout of any kind once DFU mode is entered, the only option a user has is to either wait many days for the battery to go dead or to crack open the enclosure to get access to the reset button.

I was looking forward to using the UF2 update mode provided in 0.4.0 so am disappointed I cannot move to 0.4.0 without some sort of timeout if an update is not successfully completed.

glenne:
Fixes #188 - Stuck in UF2 DFU mode forever.

Also updated README with additional guidance.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -170,9 +170,10 @@ int main(void)
 
   // Serial only mode
   bool serial_only_dfu = (NRF_POWER->GPREGRET == DFU_MAGIC_SERIAL_ONLY_RESET);
+  bool uf2_dfu = (NRF_POWER->GPREGRET == DFU_MAGIC_UF2_RESET);
 
   // start either serial, uf2 or ble
-  bool dfu_start = _ota_dfu || serial_only_dfu || (NRF_POWER->GPREGRET == DFU_MAGIC_UF2_RESET) ||
+  bool dfu_start = _ota_dfu || serial_only_dfu || uf2_dfu ||
                     (((*dbl_reset_mem) == DFU_DBL_RESET_MAGIC) && (NRF_POWER->RESETREAS & POWER_RESETREAS_RESETPIN_Msk));
 
   // Clear GPREGRET if it is our values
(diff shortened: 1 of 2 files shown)


src/main.c
```diff
@@ -251,7 +252,7 @@ int main(void)
     }
 
     // Initiate an update of the firmware.
-    if (APP_ASKS_FOR_SINGLE_TAP_RESET())
+    if (APP_ASKS_FOR_SINGLE_TAP_RESET() || uf2_dfu || serial_only_dfu)
```
hathach, line 255: spot-on both uf2 and serial_only need USB to work, it is correct to timeout if USB is not enumerated.

hathach (approved):
Thank you for the PR. It looks perfect to me. The DFU mode combination is a bit complicated for now, I will try to find my time to refactor it later on.

PR #192, opened 2021-03-02
add support for APA102 LEDs

jamesadevine:
This PR adds support for AP102 LEDs.

The board I tested this on had no LEDs, so this PR also adds a macro for conditional compilation of PWM led code too.

I tested the following configuration in `board.h`:
```c
#define LEDS_NUMBER             0
#define LED_APA102
#define LED_APA102_CLK          _PINNUM(0, 21)
#define LED_APA102_DATA         _PINNUM(0, 19)
#define LED_APA102_BRIGHTNESS   0x8
#define BOARD_RGB_BRIGHTNESS    0xffffffff
#define APA102_NUMBER           15
```

`BOARD_RGB_BRIGHTNESS` is set to max unit32_t so that colour information is not masked off. LED brightness is controlled per pixel on APA102s and relies upon the new macro `LED_APA102_BRIGHTNESS` which can be set to a maximum value of `0x1f`.

### DIFF
--- a/src/boards/boards.c
+++ b/src/boards/boards.c
@@ -24,6 +24,8 @@
 
 #include "boards.h"
 #include "nrf_pwm.h"
+#include "nrf_spim.h"
+#include "nrf_clock.h"
 #include "app_scheduler.h"
 #include "app_timer.h"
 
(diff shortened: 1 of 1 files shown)


hathach (approved):
Look good to me, thank you for the PR. Though the neopixel_init()/write() is confusing to use with dotstar ap102. I will do a rename later one, may be to rgb_init/write()

PR #193, opened 2021-03-03
Software reset

jamesadevine:
This PR adds the capability for any pin to be used as a software reset pin. This allows users to jump to an existing app with a single button press.

Depends on #192


hathach:
have to do change base back and forth to update the file diff, close-reopen doesn't seem to do the trick anymore

hathach:
hmm, user could simply do that with actual reset button. Does your board not include RST one ?

jamesadevine:
Hi @hathach 

The board I am using doesn't have a button. This PR is to enable boards that do not have a button on the reset IO, but have another button that can be used on a different pin, to jump to the application when requested.

hathach:
which board you are using, also is it an 832 or 840 ?

hathach:
close since it is out of date. it is better to just use hw reset button.

PR #194, opened 2021-03-07
Add board 'sparkfun_nrf52840_micromod'.

nitz:
Hey there!

Added the [SparkFun MicroMod nRF52840](https://www.sparkfun.com/products/16984). Opening the PR as a draft, because I don't have a PID to use. Gonna try to get in touch with someone from SparkFun and see if they've got some PIDs they'd like to use for UF2 & CircuitPython.

Hope you're well!

### DIFF
--- a/src/boards/sparkfun_nrf52840_micromod/board.h
+++ b/src/boards/sparkfun_nrf52840_micromod/board.h
@@ -59,7 +59,7 @@
 //------------- UF2 -------------//
 #define UF2_PRODUCT_NAME       "SparkFun MicroMod nRF52840"
 #define UF2_VOLUME_LABEL       "SFMM852BOOT"
-#define UF2_BOARD_ID           "micromod-nRF52840"
+#define UF2_BOARD_ID           "nRF52840-micromod-v10"
 
 #define UF2_INDEX_URL          "https://www.sparkfun.com/products/16984"
 
--- a/src/boards/sparkfun_nrf52840_micromod/board.h
+++ b/src/boards/sparkfun_nrf52840_micromod/board.h
@@ -1,0 +1,66 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2018 Ha Thach for Adafruit Industries
+ * Copyright (c) 2021 Chris Marc Dailey
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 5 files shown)


cdwilson:
FYI, here's the link to the SparkFun forum post for those following along:

https://forum.sparkfun.com/viewtopic.php?f=182&t=55251&p=223742&hilit=nrf52840+bootloader

nitz:
> FYI, here's the link to the SparkFun forum post for those following along:
> 
> https://forum.sparkfun.com/viewtopic.php?f=182&t=55251&p=223742&hilit=nrf52840+bootloader

Thanks for posting the link! I went to bed right after I posted it last night and it seems that posts (or at least from newer accounts) needed to be manually approved. 🙂

src/boards/sparkfun_nrf52840_micromod/board.h
```diff
@@ -0,0 +1,66 @@
...
+// USB
+//--------------------------------------------------------------------+
+#define USB_DESC_VID           0x1B4F
+#define USB_DESC_UF2_PID       0x0022
+#define USB_DESC_CDC_ONLY_PID  0x0023
+
+//------------- UF2 -------------//
+#define UF2_PRODUCT_NAME       "SparkFun MicroMod nRF52840"
+#define UF2_VOLUME_LABEL       "SFMM852BOOT"
+#define UF2_BOARD_ID           "micromod-nRF52840"
```
hathach, line 62: board id need to follow uf2 format https://github.com/microsoft/uf2#files-exposed-by-bootloaders 

```
The Board-ID field is machine-readable and consists of a number of dash-separated tokens. The first token is the CPU type, second is the board type, and third is the board revision. More tokens can be also added.
```

hathach (changes requested):
board id should follow std uf2 format

nitz:
Thanks, Thatch!

Done! I didn't realize that, but makes total sense.

Are you okay with the volume label, too? I felt like I was trying to cram as much info into as few characters as possible, but perhaps crammed *too* much 😂

hathach:
> Thanks, Thatch!
> 
> Done! I didn't realize that, but makes total sense.
> 
> Are you okay with the volume label, too? I felt like I was trying to cram as much info into as few characters as possible, but perhaps crammed _too_ much

to be honest, I don't use these boards, so I wouldn't care that much. Anything that makes sense is good enough.

src/boards/sparkfun_nrf52840_micromod/board.h
```diff
@@ -59,7 +59,7 @@
 //------------- UF2 -------------//
 #define UF2_PRODUCT_NAME       "SparkFun MicroMod nRF52840"
 #define UF2_VOLUME_LABEL       "SFMM852BOOT"
-#define UF2_BOARD_ID           "micromod-nRF52840"
+#define UF2_BOARD_ID           "nRF52840-micromod-v10"
```
hathach, line 62: you need to make it clear that it is `v1.0` v10 is easily mistaken as version ten. But I am fine with existing code though.

nitz: I think it *is* version 10 though. I was basing that on the schematic published on the product page, [here](https://cdn.sparkfun.com/assets/f/0/9/9/e/MicroMod_Processor_Board-nRF52840.pdf)  which has "v10" in the `REV` area.

By comparison, their similar ESP32 is [v11](https://cdn.sparkfun.com/assets/2/2/5/9/5/MicroMod_ESP32_Schematic.pdf), and the SAMD51 flavor [v12](https://cdn.sparkfun.com/assets/1/d/3/e/b/SparkFun_MicroMod_SAMD51_Processor_Board_Schematic.pdf). 

But their RP2040 one is [v1.7](https://cdn.sparkfun.com/assets/1/8/3/5/8/MicroMod-RP2040-ProcessorBoard_Schematic.pdf), and the Artemis one is [v1.6](https://cdn.sparkfun.com/assets/3/e/0/b/5/16401_MicroMod-Artemis-ProcessorBoard_schematic.pdf)

In short: Maybe those are supposed to be v1.0, v1.1, and v1.2? Should we assume it's v1.0 in this case or just match what it says on the schematic?

hathach: really, ok sorry, I normally didn't see that high number on the pcb/board world.

hathach (approved):
look good, thank you for your PR.

nitz:
Thank you! I will send a note to some sparkfun folx asking if the v10/v11/v12 are supposed to be "dot one", "dot two", etc. just in case!

hathach:
no problem, this is already merged, feel free to make follow up PR for correction later if needed.

Issue #195, opened 2021-03-09
intermittent timeouts when flashing with adafruit-nrfutil


khpeterson:
About 1/5 attempts to flash new firmware fail with a timeout (show below).  I'm running with:

feather_nrf52840_express
Adafruit_nRF52_Bootloader:  0.4.1
adafruit-nrfutil: 0.5.3.post13
osx: 10.15.7
python: 3.6 (ecdsa==0.13.2, Click==7.0, pyserial==3.4)

I have a few feather boards on hand and all exhibit the same problem as do different cables.

Upgrading target on /dev/cu.usbmodem14101 with DFU package firmware.zip. Flow control is disabled, Single bank, Touch disabled
Opened serial port /dev/cu.usbmodem14101
Starting DFU upgrade of type 4, SoftDevice size: 0, bootloader size: 0, application size: 375820
Sending DFU start packet
Sending DFU init packet
Sending firmware file
########################################
########################################
########################################
########################################
###########Timed out waiting for acknowledgement from device.

Failed to upgrade target. Error is: No data received on serial port. Not able to proceed.
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/adafruit_nrfutil-0.5.3.post13-py3.6.egg/nordicsemi/__main__.py", line 294, in serial
    dfu.dfu_send_images()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/adafruit_nrfutil-0.5.3.post13-py3.6.egg/nordicsemi/dfu/dfu.py", line 235, in dfu_send_images
    self._dfu_send_image(HexType.APPLICATION, self.manifest.application)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/adafruit_nrfutil-0.5.3.post13-py3.6.egg/nordicsemi/dfu/dfu.py", line 206, in _dfu_send_image
    self.dfu_transport.send_firmware(firmware)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/adafruit_nrfutil-0.5.3.post13-py3.6.egg/nordicsemi/dfu/dfu_transport_serial.py", line 213, in send_firmware
    self.send_packet(pkt)
[...]

Issue #162, opened 2020-08-30
Optional feature: Skip DFU with Magic


nitz:
Hiya!

I had something I added in my fork for my application, where I wanted my application to be able to skip the DFU mode (particularly the double reset wait) using a GPREGRET magic value. I've pasted the patch here if you want to look at it. It's a pretty simple thing, and I'm not sure if it's something you'd want as a feature in the mainline, but I figured I would offer it anyways! Close away if you're not interested, but just let me know if you want me to send a PR!

```diff
diff --git "a/src/main.c" "b/src/main.c"
index cf05c1a..6cd2b55 100644
--- "a/src/main.c"
+++ "b/src/main.c"
@@ -109,6 +109,7 @@ void usb_teardown(void);
 #define DFU_MAGIC_OTA_RESET             0xA8
 #define DFU_MAGIC_SERIAL_ONLY_RESET     0x4e
 #define DFU_MAGIC_UF2_RESET             0x57
+#define DFU_MAGIC_SKIP_DFU              0x6d
 
 #define DFU_DBL_RESET_MAGIC             0x5A1AD5      // SALADS
 #define DFU_DBL_RESET_APP               0x4ee5677e
@@ -175,6 +176,8 @@ int main(void)
   bool dfu_start = _ota_dfu || serial_only_dfu || (NRF_POWER->GPREGRET == DFU_MAGIC_UF2_RESET) ||
                     (((*dbl_reset_mem) == DFU_DBL_RESET_MAGIC) && (NRF_POWER->RESETREAS & POWER_RESETREAS_RESETPIN_Msk));
 
+  bool skip_dfu = NRF_POWER->GPREGRET == DFU_MAGIC_SKIP_DFU;
+
   // Clear GPREGRET if it is our values
   if (dfu_start) NRF_POWER->GPREGRET = 0;
 
@@ -198,76 +201,79 @@ int main(void)
     led_state(STATE_WRITING_FINISHED);
   }
 
-  /*------------- Determine DFU mode (Serial, OTA, FRESET or normal) -------------*/
-  // DFU button pressed
-  dfu_start  = dfu_start || button_pressed(BUTTON_DFU);
+  if (skip_dfu == false)
+  {
+    /*------------- Determine DFU mode (Serial, OTA, FRESET or normal) -------------*/
+    // DFU button pressed
+    dfu_start  = dfu_start || button_pressed(BUTTON_DFU);
 
[...]

PR #196, opened 2021-03-11
Improvement/fix of system wake up by BUTTON_DFU GPIO

lyusupov:
When an app is doing a system wake up by press and release of the BUTTON_DFU - the nRF bootloader makes unwanted entrance into DFU mode (_because of BUTTON_DFU is active_). This is not a behaviour that we all are looking for, right ?

This PR gives an option to bypass the bootloader's DFU mode by use of the MCU retention register.

This is an example of the feature usage in an Arduino sketch:

```
#define DFU_MAGIC_SKIP        0x6d

NRF_POWER->GPREGRET = DFU_MAGIC_SKIP;
pinMode(GPIO_BUTTON_DFU, INPUT_PULLUP_SENSE /* INPUT_SENSE_LOW */);

uint8_t sd_en;
(void) sd_softdevice_is_enabled(&sd_en);

if ( sd_en ) {
  sd_power_system_off();
} else {
  NRF_POWER->SYSTEMOFF = 1;
}
```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -97,15 +97,16 @@ void usb_teardown(void);
  */
 
 /* Magic that written to NRF_POWER->GPREGRET by application when it wish to go into DFU
- * - DFU_MAGIC_OTA_APPJUM used by BLEDfu service : SD is already init
- * - DFU_MAGIC_OTA_RESET entered by soft reset : SD is not init
- * - DFU_MAGIC_SERIAL_ONLY_RESET with CDC interface only
- * - DFU_MAGIC_UF2_RESET with CDC and MSC interfaces
+ * - DFU_MAGIC_OTA_APPJUM        : used by BLEDfu service, SD is already inited
+ * - DFU_MAGIC_OTA_RESET         : entered by soft reset, SD is not inited yet
+ * - DFU_MAGIC_SERIAL_ONLY_RESET : with CDC interface only
+ * - DFU_MAGIC_UF2_RESET         : with CDC and MSC interfaces
+ * - DFU_MAGIC_SKIP              : skip DFU entirely including double reset delay
  *
  * Note: for DFU_MAGIC_OTA_APPJUM Softdevice must not initialized.
  * since it is already in application. In all other case of OTA SD must be initialized
  */
-#define DFU_MAGIC_OTA_APPJUM            BOOTLOADER_DFU_START             // 0xB1
+#define DFU_MAGIC_OTA_APPJUM            BOOTLOADER_DFU_START  // 0xB1
 #define DFU_MAGIC_OTA_RESET             0xA8
 #define DFU_MAGIC_SERIAL_ONLY_RESET     0x4e
 #define DFU_MAGIC_UF2_RESET             0x57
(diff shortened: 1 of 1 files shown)


hathach:
sorry for the late response, could you tell which mcu you are using 832 or 840 ? This is indeed an improvement, though I have a plan to probably remove BUTTON_DFU on 840, and rely solely on the double-reset, this may still be needed for 832 though.

lyusupov:
>  which mcu you are using 832 or 840 ?

840

hathach (commented):
Thanks for the PR, it looks good. I think we should actually make it more generic to skip all other DFU mode and jump straight to app. This will also have another usage that allow application to perform quick-reset without waiting for double reset timeout (500ms). Let me know if you have the time to make the changes, else I will do as follow up. We are in the release windows for bootloader since I just fixed an usb issue #195

lyusupov:
> I think we should actually make it more generic to skip all other DFU mode and jump straight to app. 

The focus of this particular PR is to fix the issue while applying minimal risk to break any other of the bootloader functions.
I've also noticed a similar proposal from someone else that had been rejected in past: https://github.com/adafruit/Adafruit_nRF52_Bootloader/issues/162

> Let me know if you have the time to make the changes, else I will do as follow up. 

Please, feel yourself free to alter or extend this PR on your discretion.

hathach:
> > I think we should actually make it more generic to skip all other DFU mode and jump straight to app.
> 
> The focus of this particular PR is to fix the issue while applying minimal risk to break any other of the bootloader functions.
> I've also noticed a similar proposal from someone else that had been rejected in past: #162
> 
> > Let me know if you have the time to make the changes, else I will do as follow up.
> 
> Please, feel yourself free to alter or extend this PR on your discretion.

hmm, that is weird, now I don't see any reasons to reject that PR. I probably had a bad day at that time. I will see if I could revise and merge that one.

hathach:
I have pushed the update to skip DFU entirely and merge PR. Thank you very much for bringing this up.

lyusupov:
Thank you for this feature release!

I have one question for you to ask:
When I do:
```
NRF_POWER->GPREGRET = DFU_MAGIC_SKIP
pinMode(GPIO_BUTTON_DFU, INPUT_PULLUP_SENSE);
NRF_POWER->SYSTEMOFF = 1;
```
Will the '**double-reset**' function to operate as expected  afterwards ? _( Or the execution thread will proceed into the app at first ? )_

hathach:
with current code, it will skip the entire dfu including the double reset.

lyusupov:
This means that, to acheive entrance into DFU mode from deep sleep, it will be necessary to click the reset button **three times**:
* very first one will cause reset of the GPREGRET register:
```
if (dfu_start || dfu_skip) NRF_POWER->GPREGRET = 0;
```
* the latter two will result in regular DFU mode entrance

is that correct?

Use of three clicks instead of regular two is not too bad, but will look a bit confusing, isn't it?

hathach:
It should not since pin reset should reset the gpregret. Have you tried it out on the hw

lyusupov:
This test condition
```
  && !dfu_skip 
```
[located here](https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/master/src/main.c#L210) has no sense anymore.
We should consider to remove it sometime, right?

hathach:
yeah right, we should remove it.

PR #198, opened 2021-03-30
adding support for generic nrf52833 board

braingram:
I had a need to get this working on a custom nrf52833 board and wanted to share the result. I don't have a pca10100 to test with but could get one if it would prove useful in validating this pull request. Otherwise I added a 'generic nrf52833'.

I wasn't sure what to put for usb vendor and product ids so please do let me know if those should be changed.

### DIFF
--- a/src/boards/generic_nrf52833/board.h
+++ b/src/boards/generic_nrf52833/board.h
@@ -1,0 +1,60 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2018 Ha Thach for Adafruit Industries
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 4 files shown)


hathach (changes requested):
Thank for your PR. The changes look good to me. **Though we won't accept generic board**. But you don't have the buy pca10010 to submit PR, just update your board.h with correct LED, BUTTON. There aren't much to configure anyway. If there is issue with gpio, people could just file issue for that.

For VID, PID, it is mandatory by USB-IF to assigned unique one for each board/product. Since you are adding pca10010 and `VID = 0x1915`, it is best to ask this on Nordic forum, with link to this PR, ask them to see which PID you could use.

braingram:
Thanks for the feedback. The denial of the generic board totally makes sense to me. I have a pca10100 on order to test out the bootloader on that device. Once it's in I'll change this pull request to focus only on changes to the existing pca10100. It's likely the only remaining change will be to the src/usb/uf2/uf2cfg.h file.

jpconstantineau:
@braingram Some of the changes were merged in with #251 I suggest reviewing your changes (see comments in my PR) and decide whether to close or update this PR

PR #205, opened 2021-05-30
Adding EByte E104-BT5032A module

jpconstantineau:
This will allow to re-use the "Serial-AT" module and be able to reprogram it.


hathach (approved):
look good, thanks for the PRs

PR #206, opened 2021-05-30
EByte-E104: fix to LED direction

jpconstantineau:
Sorry about this simple PR.  The LED direction was upside down in my original PR


hathach (approved):
no problems at all

PR #208, opened 2021-06-02
Add support for nice!nano v2

Nicell:
The second version of the nice!nano has the same pinout with slight changes to the power system requiring for the `REGOUT0` UICR value to be set. This PR allows for the nice!nano board definition to support both versions.


hathach (approved):
look good, thanks for the pr

PR #209, opened 2021-06-13
Add makerdiary m60 keyboard (uses nrf52840_m2)

y4m-y4m:
This PR adds the [Makerdiary M60 Mechanical Keyboard](https://wiki.makerdiary.com/m60/). The keyboard uses the Makerdiary nRF52840 M.2 module. 
I've been using this keyboard for a while with the nrf52840_m2 bootloader. It worked, but there was no way to get into bootloader mode via button, because the buttons defined are for the Makerdiary M.2 Developer Kit. The keyboard has only one button on pin 0_27, the second defined button is the one from nrf52840_m2. I tested the bootloader and I'm able to access bootloader mode via the button on the back now.
I'm not sure what i should use for the PID or VID, for now these are the ones from nrf52840_m2.

### DIFF
--- a/.github/workflows/githubci.yml
+++ b/.github/workflows/githubci.yml
@@ -36,6 +36,7 @@ jobs:
           - 'ebyte_e73_tbb'
           - 'electronut_labs_papyr'
           - 'ikigaisense_vita'
+          - 'm60_keyboard'
           - 'mdk_nrf52840_dongle'
           - 'nice_nano'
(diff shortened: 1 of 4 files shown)


hathach (approved):
Thank you for your PR. Everything look good, for VID/PID, since it use the m2 module, I think it is OK to re-use the PID

y4m-y4m:
Added to ci. Thanks.

Issue #20, opened 2018-08-14
BLE OTA is not run with SD 6.1.0 


hathach:
The same code/hex run fine with SD 6.0.0.  Have tried to troubleshoot without any clues. 
Posted in Nordic devzone for help
https://devzone.nordicsemi.com/f/nordic-q-a/37473/advertising-does-not-start-with-s140-6-1-0-same-hex-run-without-issue-with-s140-6-0-0

hathach:
sum up: adv buffer, and gap_adv must live long enough until advertising is complete

Issue #16, opened 2018-08-09
return bootloader version for nrfutil


hathach:
need to modify nrfutil as well.

hathach:
closed since this can be done simply with uf2 info file

PR #211, opened 2021-06-26
GCC 11 Build fixes

DuMaM:
Fixed build errors for GCC11

``` c
src/main.c:456:5: error: 'soc_evt' may be used uninitialized [-Werror=maybe-uninitialized]
  456 |     pstorage_sys_event_handler(soc_evt);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/main.c: In function 'proc_ble':
src/main.c:425:24: error: '*(ble_evt_t *)(&ev_buf[0]).header.evt_id' may be used uninitialized [-Werror=maybe-uninitialized]
  425 |     switch (evt->header.evt_id)
      |             ~~~~~~~~~~~^~~~~~~

lib/sdk11/components/ble/ble_services/ble_dfu/ble_dfu.c: In function 'ble_dfu_init':
lib/sdk11/components/ble/ble_services/ble_dfu/ble_dfu.c:487:36: error: 'service_uuid.type' may be used uninitialized [-Werror=maybe-uninitialized]
  487 |     p_dfu->uuid_type = service_uuid.type;
      |                        ~~~~~~~~~~~~^~~~~
```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -416,7 +416,7 @@ uint32_t proc_ble(void)
 
   // Init header
   ble_evt_t* evt = (ble_evt_t*) ev_buf;
-  evt->header.evt_id = 0;
+  evt->header.evt_id = BLE_EVT_INVALID;
 
   // Get BLE Event
   uint32_t err = sd_ble_evt_get(ev_buf, &ev_len);
--- a/src/main.c
+++ b/src/main.c
@@ -411,7 +411,7 @@ void assert_nrf_callback (uint16_t line_num, uint8_t const * p_file_name)
 // Process BLE event from SD
 uint32_t proc_ble(void)
 {
-  __ALIGN(4) uint8_t ev_buf[ BLE_EVT_LEN_MAX(BLEGATT_ATT_MTU_MAX) ];
+  __ALIGN(4) uint8_t ev_buf[ BLE_EVT_LEN_MAX(BLEGATT_ATT_MTU_MAX) ] = {0};
   uint16_t ev_len = BLE_EVT_LEN_MAX(BLEGATT_ATT_MTU_MAX);
 
   // Get BLE Event
@@ -448,7 +448,7 @@ uint32_t proc_ble(void)
 // process SOC event from SD
 uint32_t proc_soc(void)
 {
-  uint32_t soc_evt;
+  uint32_t soc_evt = 0;
   uint32_t err = sd_evt_get(&soc_evt);
 
   if (NRF_SUCCESS == err)
--- a/src/main.c
+++ b/src/main.c
@@ -411,8 +411,12 @@ void assert_nrf_callback (uint16_t line_num, uint8_t const * p_file_name)
 // Process BLE event from SD
 uint32_t proc_ble(void)
 {
-  __ALIGN(4) uint8_t ev_buf[ BLE_EVT_LEN_MAX(BLEGATT_ATT_MTU_MAX) ] = {0};
+  __ALIGN(4) uint8_t ev_buf[ BLE_EVT_LEN_MAX(BLEGATT_ATT_MTU_MAX) ];
   uint16_t ev_len = BLE_EVT_LEN_MAX(BLEGATT_ATT_MTU_MAX);
+
+  // Init header
+  ble_evt_t* evt = (ble_evt_t*) ev_buf;
+  evt->header.evt_id = 0;
 
   // Get BLE Event
   uint32_t err = sd_ble_evt_get(ev_buf, &ev_len);
@@ -420,8 +424,6 @@ uint32_t proc_ble(void)
   // Handle valid event, ignore error
   if( NRF_SUCCESS == err)
   {
-    ble_evt_t* evt = (ble_evt_t*) ev_buf;
-
     switch (evt->header.evt_id)
     {
       case BLE_GAP_EVT_CONNECTED:
--- a/lib/sdk11/components/ble/ble_services/ble_dfu/ble_dfu.c
+++ b/lib/sdk11/components/ble/ble_services/ble_dfu/ble_dfu.c
@@ -463,7 +463,7 @@ uint32_t ble_dfu_init(ble_dfu_t * p_dfu, ble_dfu_init_t * p_dfu_init)
 
     p_dfu->conn_handle = BLE_CONN_HANDLE_INVALID;
 
-    ble_uuid_t service_uuid;
+    ble_uuid_t service_uuid = {0};
     uint32_t   err_code;
 
     const ble_uuid128_t base_uuid128 =


DuMaM:
I also fixed build for one submodule
https://github.com/hathach/tinyusb/pull/923

src/main.c
```diff
@@ -411,7 +411,7 @@ void assert_nrf_callback (uint16_t line_num, uint8_t const * p_file_name)
 // Process BLE event from SD
 uint32_t proc_ble(void)
 {
-  __ALIGN(4) uint8_t ev_buf[ BLE_EVT_LEN_MAX(BLEGATT_ATT_MTU_MAX) ];
+  __ALIGN(4) uint8_t ev_buf[ BLE_EVT_LEN_MAX(BLEGATT_ATT_MTU_MAX) ] = {0};
```
hathach, line 414: It is not worth the effort to initialize the whole buffer, there is error check for sd_ble_evt_get() already. We should fine another way to keep gcc happy.

hathach (changes requested):
All variables are used as output of function with proper error check and it does not worth effort to initialize whole buffer this way. We need to find a better way to make gcc happy

src/main.c
dhalbert: How about using pragma, if you are confident it is mistaken?
```c
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
...  // ignore the diagnostic for code here
#pragma GCC diagnostic pop
```

DuMaM: @dhalbert It's also good idea, but I don't like using a per-preprocessor directives and makes it compiler specific.
Right now I just initialized header parameter directly. Maybe it's not a cleanest solution, but right now i don't have any other ideas.

DuMaM:
Anyway i noticed that something is wrong with bootloader itself when i complie it with gcc11.
I dont know why I'm unable to flash any image and boot to them.
In both uf2 and after flashing with JTAG...

src/main.c
```diff
@@ -414,14 +414,16 @@ uint32_t proc_ble(void)
   __ALIGN(4) uint8_t ev_buf[ BLE_EVT_LEN_MAX(BLEGATT_ATT_MTU_MAX) ];
   uint16_t ev_len = BLE_EVT_LEN_MAX(BLEGATT_ATT_MTU_MAX);
 
+  // Init header
+  ble_evt_t* evt = (ble_evt_t*) ev_buf;
+  evt->header.evt_id = 0;
```
hathach, line 419: 0 is correct value, though, could you use the `BLE_EVT_INVALID`  instead, which better reflect the invalid event ID.

hathach (changes requested):
init the event id is spot-on, though using `BLE_EVT_INVALID` instead of 0, would be more descriptive if you don't mind changing it.

hathach:
> Anyway i noticed that something is wrong with bootloader itself when i complie it with gcc11.
> I dont know why I'm unable to flash any image and boot to them.
> In both uf2 and after flashing with JTAG...
> 
> One I took from Github Releases works without issues

which command you use for flashing, also which board and setup you are using ? The release is built with gcc 9.3.1 https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/master/.github/workflows/githubci.yml#L71 Since I run into an issue with one of later gcc and haven't got time to address it since. It may have something with optimization etc with latest gcc, feel free to open an separate issue with more info. We can easier address it there, since this PR will be closed when merged.

DuMaM:
> > Anyway i noticed that something is wrong with bootloader itself when i complie it with gcc11.
> > I dont know why I'm unable to flash any image and boot to them.
> > In both uf2 and after flashing with JTAG...
> > One I took from Github Releases works without issues
> 
> which command you use for flashing, also which board and setup you are using ? The release is built with gcc 9.3.1 https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/master/.github/workflows/githubci.yml#L71 Since I run into an issue with one of later gcc and haven't got time to address it since. It may have something with optimization etc with latest gcc, feel free to open an separate issue with more info. We can easier address it there, since this PR will be closed when merged.

So the same command works for release version and do not work for GCC11

# Build
``` sh
make BOARD=particle_xenon all
...
...
...
LD particle_xenon_bootloader-0.6.0-2-g6e33825-dirty.out
   text    data     bss     dec     hex filename
  31992    1576   22578   56146    db52 _build/build-particle_xenon/particle_xenon_bootloader-0.6.0-2-g6e33825-dirty.out
Create particle_xenon_bootloader-0.6.0-2-g6e33825-dirty.hex
Create particle_xenon_bootloader-0.6.0-2-g6e33825-dirty_nosd.hex
Create update-particle_xenon_bootloader-0.6.0-2-g6e33825-dirty_nosd.uf2
Converting to uf2, output size: 73728, start address: 0x0
Wrote 73728 bytes to _build/build-particle_xenon/update-particle_xenon_bootloader-0.6.0-2-g6e33825-dirty_nosd.uf2
Create particle_xenon_bootloader-0.6.0-2-g6e33825-dirty_s140_6.1.1.hex
Zip created at _build/build-particle_xenon/particle_xenon_bootloader-0.6.0-2-g6e33825-dirty_s140_6.1.1.zip
```

# Flashing
``` bash
#!/bin/bash

# https://docs.particle.io/tutorials/learn-more/xenon-circuit-python/

FIRMWARE=$(find . -name "particle_xenon*s140_*.hex")
echo "Flashing $FIRMWARE"

openocd -f interface/ftdi/jtag-lock-pick_tiny_2.cfg \
        -c "transport select swd" \
        -f target/nrf52.cfg \
        -c "init; reset; halt" \
        -c "nrf5 mass_erase" \
        -c "program $FIRMWARE 0x00000000 verify reset" \
        -c "exit"
```

TOOLS VERSION
``` sh
Using built-in specs.
COLLECT_GCC=arm-none-eabi-gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/arm-none-eabi/11.1.0/lto-wrapper
Target: arm-none-eabi
Configured with: /build/arm-none-eabi-gcc/src/gcc-11.1.0/configure --target=arm-none-eabi --prefix=/usr --with-sysroot=/usr/arm-none-eabi --with-native-system-header-dir=/include --libexecdir=/usr/lib --enable-languages=c,c++ --enable-plugins --disable-decimal-float --disable-libffi --disable-libgomp --disable-libmudflap --disable-libquadmath --disable-libssp --disable-libstdcxx-pch --disable-nls --disable-shared --disable-threads --disable-tls --with-gnu-as --with-gnu-ld --with-system-zlib --with-newlib --with-headers=/usr/arm-none-eabi/include --with-python-dir=share/gcc-arm-none-eabi --with-gmp --with-mpfr --with-mpc --with-isl --with-libelf --enable-gnu-indirect-function --with-host-libstdcxx='-static-libgcc -Wl,-Bstatic,-lstdc++,-Bdynamic -lm' --with-pkgversion='Arch Repository' --with-bugurl=https://bugs.archlinux.org/ --with-multilib-list=rmprofile
Thread model: single
Supported LTO compression algorithms: zlib zstd
gcc version 11.1.0 (Arch Repository) 

arm-none-eabi-newlib-4.1.0
arm-none-eabi-binutils-2.36.1
```

hathach (approved):
changes look good, thank you for your PR

hathach:
@DuMaM latest tinyusb could have its own breaking changes. It is best for you to stick to the version of the sub-modules.  You could cherry-pick or have a modification to suppress the warning instead of bump it up to master. Once after a while, I will do bump up tinyusb when I think is needed since not all changes is needed for this particular bootloader.

src/main.c
DuMaM: done

daniel-thompson:
Sorry for poking an old thread but was the bootloader *tested* after being compiled with gcc-11 or was it just build tested?

The reason I ask is that I've been looking into some broken codegen with gcc-11 and concluded that the data is *not* actually uninitialized meaning the changes to suppress the warnings end up hiding a real codegen problem.

The Nordic Softdevice SDK headers contain broken inline assembler contraints and, when combined with inter-proceedural optimizations, the code gen breaks as "dead" code (that is not actually dead) is striped from the resulting binaries. The result of these broken assembler contraints is that the compiler *incorrectly* concludes some of the arguments to functions calls that used the SVCALL() macro are unused. This doesn't just results in the uninitialized data warnings that this PR papers over. It also causes the call sites not to bother setting up the arguments in the first place. The result of this is that the bootloader makes svc calls to the SoftDevice without initializing the arguments.

At a codegen level it looks like this:
~~~
Dump of assembler code for function softdev_mbr_init:
   0x000f4788 <+0>:     sub     sp, #16
   0x000f478a <+2>:     mov     r0, sp
   0x000f478c <+4>:     add     sp, #16
   0x000f478e <+6>:     b.w     0xf4758 <sd_mbr_command>
~~~

Naturally the best fix is to fix the SVCALL() macro but this is pretty complex (because I can't find a way to fix it without changing all the sites that use the macro).  However you can turn off one of the inter-procedural optimization paths. That has the effect of fixing the warnings (without any extra initialization) *and* keeping the codegen the same as gcc-10.

In other words, with -fno-ipa-modref, then the codegen is fixed:
~~~
(gdb) disass softdev_mbr_init 
Dump of assembler code for function softdev_mbr_init:
   0x000f4788 <+0>:     push    {r0, r1, r2, r3, r4, lr}
   0x000f478a <+2>:     movs    r3, #0
   0x000f478c <+4>:     strd    r3, r3, [sp, #4]
   0x000f4790 <+8>:     str     r3, [sp, #12]
   0x000f4792 <+10>:    mov     r0, sp
   0x000f4794 <+12>:    movs    r3, #2
   0x000f4796 <+14>:    str     r3, [sp, #0]
   0x000f4798 <+16>:    bl      0xf4758 <sd_mbr_command>
   0x000f479c <+20>:    add     sp, #20
   0x000f479e <+22>:    ldr.w   pc, [sp], #4
~~~

Obviously a proper fix would fix the inline assembler contraints in the SVCALL() macro. Unfortunately this requires restrucuting all the call sites (because to get the contraints correct we must know the number and names of the function arguments). Thus for now I have been making a simpler change (with also fixes all the uninitialized variable warnings):
~~~
diff --git a/Makefile b/Makefile
index 238dbb9..e2d5fc4 100644
--- a/Makefile
+++ b/Makefile
@@ -270,6 +270,16 @@ CFLAGS += -Wno-unused-parameter -Wno-expansion-to-defined
 # TinyUSB tusb_hal_nrf_power_event
 CFLAGS += -Wno-cast-function-type
 
+# Nordic Softdevice SDK header files contain inline assembler that has^M
+# broken contraints. As a result the IPA-modref pass is able to "prove"^M
+# that arguments to wrapper functions generated with the SVCALL() macro^M
+# are unused and, as a result, the optimizer removes the code in the callers^M
+# that sets up these arguments (which result in broken syscalls). Since^M
+# the broken headers come from Nordic-supplied zip files and are not^M
+# easily patched then, for now, we'll simply disable some of the^M
+# gcc-11 inter-procedural optimizations.^M
+CFLAGS += -fno-ipa-modref^M
+^M
 # Defined Symbol (MACROS)
 CFLAGS += -D__HEAP_SIZE=0
 CFLAGS += -DCONFIG_GPIO_AS_PINRESET
~~~

I'm happy to submit a PR if this approach to fixing the codegen is acceptable as a temporary workaround.

hathach:
@daniel-thompson **Accepted !!** it is great workaround if we could get SVCALL to work with latest gcc. It has been a puzzle to me and couldn't find time to investigate it. The gcc is fixed with 9.3.1 for ci testing and generating binaries https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/master/.github/workflows/githubci.yml#L74

Please make an PR, I will try to test it out with my actual hardware

PR #214, opened 2021-08-04
Fix flash with more than 512kb

### LINKED ISSUE #213: Unable to flash UF2 where flash is >512kb
**Set up**
Feather nRF52840 Express with 6.0.0 bootloader. CircuitPython e9369d50e16e557ccd1a98c06d6e36bb87b908b7 . UF2s for the opt build flash ok but DEBUG=1 builds do not.

**Describe the bug**

Only part of the >512k UF2 is flashed leaving the higher end of the flash incorrect. In CircuitPython this means the .data section is not correct and assertions in DEBUG=1 mode fail as a result.

**To Reproduce**

1. Enter the bootloader
2. Drag on
[firmware-debug.uf2.zip](https://github.com/adafruit/Adafruit_nRF52_Bootloader/files/6919834/firmware-debug.uf2.zip) after unzipping.
3. Red LED flashes quickly and board doesn't restart.
4. Manual restart leads to CP assertion failure.

Here is an opt build that works: [firmware-opt.uf2.zip](https://github.com/adafruit/Adafruit_nRF52_Bootloader/files/6919841/firmware-opt.uf2.zip)

Note that the broken UF2 is >1MB in size.

hathach:
fix #213 

```
Write addr = 0x000ACD00, block = 2142 (2142 of 2191)
Write addr = 0x000ACE00, block = 2143 (2143 of 2191)
Write addr = 0x000ACF00, block = 2144 (2144 of 2191)
```

I found the issue, the bootloader is limited its size to 512KB since it reserved the rest for internal file system ( for all boards), since the very first implementation of the bootloader.
- For Arduino, all boards is served with 28KB for storing bonding information. 
- For Circuitpython, there seems to be 3 section FLASH_BLE_CONFIG (32KB), FLASH_NVM (8KB), FLASH_FATFS (can be 0 on qspi flash dev).

For that I think we will increase reserved size to 40KB. The overflow to fatfs should be checked by circuitpyhon  linker, just try DEBUG=1 with boards internal flash, it overflows and won't linked which is good.

_Originally posted by @hathach in https://github.com/adafruit/Adafruit_nRF52_Bootloader/issues/213#issuecomment-892501613_

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,14 +104,17 @@ ifeq ($(MCU_SUB_VARIANT),nrf52)
   SD_NAME = s132
   DFU_DEV_REV = 0xADAF
   CFLAGS += -DNRF52 -DNRF52832_XXAA -DS132
+  CFLAGS += -DDFU_APP_DATA_RESERVED=7*4096
 else ifeq ($(MCU_SUB_VARIANT),nrf52833)
   SD_NAME = s140
   DFU_DEV_REV = 52840
   CFLAGS += -DNRF52833_XXAA -DS140
+  CFLAGS += -DDFU_APP_DATA_RESERVED=7*4096
 else ifeq ($(MCU_SUB_VARIANT),nrf52840)
   SD_NAME = s140
   DFU_DEV_REV = 52840
   CFLAGS += -DNRF52840_XXAA -DS140
+  CFLAGS += -DDFU_APP_DATA_RESERVED=10*4096
 else
   $(error Sub Variant $(MCU_SUB_VARIANT) is unknown)
 endif
@@ -232,15 +235,6 @@ IPATH += $(SD_PATH)/$(SD_FILENAME)_API/include/nrf52
 # Compiler Flags
 #------------------------------------------------------------------------------
 
-# Debug option use RTT for printf
-ifeq ($(DEBUG), 1)
-	RTT_SRC = lib/SEGGER_RTT
-	
-	CFLAGS += -DCFG_DEBUG -DSEGGER_RTT_MODE_DEFAULT=SEGGER_RTT_MODE_BLOCK_IF_FIFO_FULL
-	IPATH += $(RTT_SRC)/RTT
-  C_SRC += $(RTT_SRC)/RTT/SEGGER_RTT.c
-endif
-
 #flags common to all targets
 CFLAGS += \
 	-mthumb \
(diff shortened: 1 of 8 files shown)


tannewt:
I'd expect the bootloader to only protect itself. Protecting user data is the job of the UF2 when it specifies what blocks to write. We've actually taken advantage of this behavior on samd because current.uf2 will include the fatfs data and make it possible to share it.

The CP linker script should already fail when not enough room is left after the fatfs partition is set aside.

hathach:
yeah, you are right, but it is a bit more complicated and originally from existing nordic sdk code with dual bank (which our first version based on). And It is expanded to allow in-place update bootloader + sd without touching the user area region and/or avoid to write to user firmware if possible by using the highest flash as temporarily until whole image is receive.  https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/master/src/usb/uf2/ghostfat.c#L449 

samd upgrade is a bit different, by having the user application to write/update bootloader flash, then destroy itself to prevent re-updating. Which is more generic (which is currently used by tinyuf2 for most ports).

We can of course just define it as 0x0 and leave that to application linker, though it could take more effort for refactoring & testing. Which we could do it later, I could open an issue for that If you insist.

tannewt:
What does this have to do with updating the bootloader? This should be a simpler case of updating a user application. What is the purpose of protecting app data? I'd only expect the bootloader and MBR to be protected.

Am I correct thinking it was partially flashing before failing? That would explain how part of our image worked but the end of it wasn't included. Perhaps the flash region needs to be validated at the start before any erase.

hathach:
> What does this have to do with updating the bootloader? This should be a simpler case of updating a user application. What is the purpose of protecting app data? I'd only expect the bootloader and MBR to be protected.

For this issue, it doesn't. but I just realized, circuipython use 40KB for bonding and nvm (user data) more than currently reserved 28KB. Therefore bootloader needs to adjust to increase that for not overwriting those region when updating itself. The reason is mentioned in above comment.

> Am I correct thinking it was partially flashing before failing? That would explain how part of our image worked but the end of it wasn't included. Perhaps the flash region needs to be validated at the start before any erase.

Yes, it failed when it goes over 512 KB, which is fixed now, originally, it is the capped limit of user application due to internal filesystem when the very first port of circuitpython for nrf52840 and nrf52832. It is updated to remove that in https://github.com/adafruit/Adafruit_nRF52_Bootloader/pull/214/files#diff-d7b2438df61e4c16edd0db62cd50ddddb0e72b6984050d42ea9d7ae0bb94738bR21

Yes, we can pre-calculate the end address based on the start + total block to actively reject the uf2 instead of writing part of it. I just opened an issue for this, will fix it later on #215

tannewt:
Updating the bootloader is pretty rare. I think it's ok to wipe user data in that process. Other bootloaders don't make any promise about it.

Thanks for opening #215!

PR #219, opened 2021-08-29
glasses def

ladyada:
new bootloader for board!


hathach (approved):
look good, i will update the url with PID when it is available later on.

PR #221, opened 2021-09-07
Use -fno-ipa-modref when building with gcc-11 (or later)

daniel-thompson:
As the large comment in the body of the patch, there are mistakes in
the inline assembler fragments contained in the Nordic Softdevice header
files. Fixing the problem is a bit fiddly so for now we'll simply work
around the issue by disabling an optimization pass that was introduced
in gcc-11.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -270,6 +270,18 @@ CFLAGS += -Wno-unused-parameter -Wno-expansion-to-defined
 # TinyUSB tusb_hal_nrf_power_event
 CFLAGS += -Wno-cast-function-type
 
+# Nordic Softdevice SDK header files contains inline assembler that has
+# broken constraints. As a result the IPA-modref pass, introduced in gcc-11,
+# is able to "prove" that arguments to wrapper functions generated with
+# the SVCALL() macro are unused and, as a result, the optimizer will remove
+# code within the callers that sets up these arguments (which results in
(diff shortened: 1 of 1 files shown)


daniel-thompson:
This is the PR discussed in #211 ...

Note that this PR does not currently revert the structure initializations added in #211 although I think it could because those warnings were arguably correct!

tannewt:
Is there an issue on a nordic repo to fix the assembly that is broken by this? I had to add this flag for CircuitPython to work as well.

daniel-thompson:
> Is there an issue on a nordic repo to fix the assembly that is broken by this? I had to add this flag for CircuitPython to work as well.

Yes, the inline assembler can be fixed but fixing it properly is a fairly big change since it would require modifying *all* the callers of SVCALL() (IIRC S132 alone has 138) so they can provide the number and names of all the arguments that the inline assembler will consume. I thought it unlikely any project would be keen to carry around such a pervasive change.

I did try a couple of different hacks with weak symbols or aliases to try keep SVCALL() unmodified but they all had drawbacks (linker can't garbage collect weak symbols, can't work header only, etc, etc).

hathach (approved):
thank you very much for the PR. I actually haven't tested this since ARM only officially release 10.3  so far. However since it is tested in https://github.com/adafruit/circuitpython/pull/5319. It is good to go.

PR #226, opened 2021-10-07
Add New Board: BlueMicro833

jpconstantineau:
Adding a nRF52833 board with the form factor of a Pro Micro that uses the EByte E73-2G4M08S1E Module.

Waiting on approval of PID request at openmoko:
https://github.com/openmoko/openmoko-usb-oui/pull/34

If anyone is interested:
Hardware design files: https://github.com/jpconstantineau/BlueMicro833_hardware

### DIFF
--- a/.github/workflows/githubci.yml
+++ b/.github/workflows/githubci.yml
@@ -32,6 +32,7 @@ jobs:
           - 'arcade_feather_nrf52840_express'
           - 'arduino_nano_33_ble'
           - 'bast_ble'
+          - 'bluemicro_nrf52833'
           - 'bluemicro_nrf52840'
           - 'ebyte_e104_bt5032a'
           - 'ebyte_e73_tbb'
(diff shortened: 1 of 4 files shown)


carrino:
I'm interested in getting nRF52833 working as well.

It looks like the Flash and RAM are both wrong.

```
  204, 0x100000,                                // FLASH_BYTES = 0x100000
  205, 0x40000,                                 // RAM_BYTES = 0x40000
```

These should both be halved.  

I am interested to know what testing was done because my understanding from the other PRs is that more than 4 files need to change to get nRF52833 working.

jpconstantineau:
That's a good question.  I took the config from the PCA10100 [here](https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/master/src/boards/pca10100/pinconfig.c#L11) and used that.  Since they are both nrf52833, I assumed that they would be good.

@hathach perhaps you can comment.
@dhalbert would this cause the issue with Circuitpython on the nrf52833?

PR #231, opened 2021-11-13
add 833 support for v7

kbladewht:


src/boards/keystar/board.h
```diff
@@ -0,0 +1,65 @@
...
+#define BLEDIS_MODEL           "Orion-OTA"
+
+//--------------------------------------------------------------------+
+// USB
+//--------------------------------------------------------------------+
+
+// Shared VID/PID with Feather nRF52840, will be disabled for building in the future
+#define USB_DESC_VID           0x239A
+#define USB_DESC_UF2_PID       0x0029
+#define USB_DESC_CDC_ONLY_PID  0x0029
```
hathach, line 58: The PID is already used by Feather nRF52840, each board must have its unique vid/pid. If you don't have one, try to apply for one https://pid.codes/

src/boards/nrf52833/board.h
```diff
@@ -0,0 +1,65 @@
...
+
+//--------------------------------------------------------------------+
+// USB
+//--------------------------------------------------------------------+
+
+// Shared VID/PID with Feather nRF52840, will be disabled for building in the future
+#define USB_DESC_VID           0x239A
+#define USB_DESC_UF2_PID       0x0029
+#define USB_DESC_CDC_ONLY_PID  0x0029
+#define UF2_VOLUME_LABEL   "KSTARORION833"
```
hathach, line 59: same as above, each board must have unique vid/pid

src/boards/nrf52833/board.h
```diff
@@ -0,0 +1,65 @@
+/*
```
hathach, line 1: which board is this nrf52833, it is to generic, maybe adding an manufacturer name to distinguish it from other 833 boards.

hathach (changes requested):
Thank you for the PR. It looks OK, however, per USB specs, we must enforce an unique vid/pid for each board. The vid/pid cannot be re-used. You need to apply for 1 from donor as mentioned in below reviews.

jpconstantineau:
@kbladewht Do you think you will be able to work on this? 
When referring to #246, your work would benefit all nRF52833 users and the community would really appreciate if the new board you have could be split from this PR and the work you mention in the title of the PR (support for SD V7) be merged in.  I suspect that the blocker is indeed that USB VID/PID. Splitting this PR in two would remove the blocker for merging the core of this effort. 

Thanks

hathach:
@jpconstantineau OP seems to be busy atm, if you need this asap, you could check out OP's branch, make update yourself then submit it as PR. Make sure the commmit log is as it is that will give credit to OP. When your PR is merged, github is smart enough to set this PR status to merge as well.

jpconstantineau:
@hathach Thanks for the comments.  I have been busy looking at the circuitpython nrf52833 issues and am close to move towards this.  Good to know that github is clever enough to close two PRs if one is the continuation of the other.

PR #248, opened 2021-12-10
add 833 support for v7 Softdevice (while not including new boards)

jpconstantineau:
This PR brings the work of @kbladewht  but removes the 3 new boards that prevented this work from being merged in.
This should also close #231


jpconstantineau:
Let me clean this up and rebase this first...

PR #250, opened 2021-12-10
add 833 support for v7 Softdevice (while not including new boards)

jpconstantineau:
This PR brings the work of @kbladewht but removes the 3 new boards that prevented this work from being merged in.
This should also close #231

Did a rebase off of the head of the repo.  Closed the previous PRs as they brought in unwanted commits.  This one is in line with the original PR #231 but removes the additional boards.

It also does not change the config of the existing boards to use the new SD.  This should come separate in another commit once this one is in.


Makefile
```diff
@@ -15,7 +15,10 @@ TUSB_PATH    = lib/tinyusb/src
 NRFX_PATH    = lib/nrfx
 SD_PATH      = lib/softdevice/$(SD_FILENAME)
 
-SD_VERSION   = 6.1.1
+ifndef SD_VERSION
+   SD_VERSION   = 6.1.1
+endif
+
```
hathach, line 21: where would the `SD_VERSION` is defined for 833 ?

jpconstantineau: Let me provide an example for the bluemicro833.

hathach: I see, it is defined by board.mk, could you update the pca10100 as well.

jpconstantineau: Yes.  I will also be able to test it with CircuitPython as well.

jpconstantineau: @hathach I have a question for you. I asked @tannewt on one of his livestream and he indicated that you would be the one to ask...  The makefile for CP (nrf - see [here](https://github.com/adafruit/circuitpython/blob/main/ports/nrf/Makefile#L282)) uses a specific command to flash a bit to indicate that the application checks out and to bypass the CRC check.  Do you know if setting this bit should be done by the application UF2 or the bootloader/softdevice UF2?  Is this something new with SD7.x? Or was it there all along and the bit is set for the SD in this repo?

jpconstantineau: @hathach you can disregard my question regarding the CRC/app check bit.  When using the 7.x booloader, all I need to do is drop the UF2 and it works perfectly.

CircuitPython loads fine with the latest build as one would expect.

hathach (approved):
look great now, thank you for your PR

PR #256, opened 2021-12-22
Use separate UF2_FAMILY for nRF52833

xudongzheng:
Per my understanding, the logic in `src/usb/uf2/ghostfat.c` will prevent nRF52840 UF2 firmware from being flashed to nRF52833 and vice versa (so long as `uf2conv.py` is invoked with the correct UF2_FAMILY token).

### DIFF
--- a/src/boards/bluemicro_nrf52833/pinconfig.c
+++ b/src/boards/bluemicro_nrf52833/pinconfig.c
@@ -11,7 +11,7 @@ const uint32_t bootloaderConfig[] =
   204, 0x80000,                                 // FLASH_BYTES
   205, 0x20000,                                 // RAM_BYTES
   208, (USB_DESC_VID << 16) | USB_DESC_UF2_PID, // BOOTLOADER_BOARD_ID = USB VID+PID, used for verification when updating bootloader via uf2
-  209, 0xada52840,                              // UF2_FAMILY = 0xada52840
+  209, 0x621e937a,                              // UF2_FAMILY = 0x621e937a
   210, 0x20,                                    // PINS_PORT_SIZE = PA_32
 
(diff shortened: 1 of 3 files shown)


jpconstantineau:
Will this have an impact on circuitpython UF2 builds?
I suspect they use the same family.
There are no Bluemicro833 in the wild but I am not sure about the pca10100.
Is this related to ZMK or other firmwares?

jpconstantineau:
If this is pushed in, changes to the CircuitPython builds will be needed.  See makefile [here](https://github.com/adafruit/circuitpython/blob/main/ports/nrf/Makefile#L267)

xudongzheng:
I think there would be a case for keeping 0xADA52840 if there are lots of nRF52833 boards with the existing bootloader out there.

Wouldn't CircuitPython already be building different firmware for nRF52833 compared to nRF52840? I'm not too familiar with it. Is the Python script also deployed via UF2?

hathach:
yeah, it should use different uf2 family to prevent incompatible updatring, before merging this PR. you need to submit an PR to uf2 specs to update https://github.com/microsoft/uf2/blob/master/utils/uf2families.json .

Note: I am not sure if uf2 maintainer would agree to use `ada52833`, if not you could just generate a random number.

jpconstantineau:
I agree we should get a new number.  Now is the time to make it right and get the changes done here and at other places (CP).  There aren't too many of these nrf52833 out there yet.  And those people who have them would more likely have a dev kit where they can update the bootloader easily.

xudongzheng:
I've changed the UF2_FAMILY to 0x621e937a.

hathach (approved):
look good, thank you for your PR

PR #258, opened 2021-12-29
Add New Board: nRF52840-BBoard

ddB0515:
Adding nRF52840-bboard

More info can be found here
https://github.com/ddB0515/nRF52840-BBoard


.github/workflows/githubci.yml
```diff
@@ -43,6 +43,7 @@ jobs:
           - 'mikoto'
           - 'nice_nano'
           - 'nrf52840_m2'
+          - 'nrf52840_bboard'
```
hathach, line 46: should be 1 line up

src/boards/nrf52840_bboard/board.h
```diff
@@ -0,0 +1,68 @@
...
+//--------------------------------------------------------------------+
+#define BLEDIS_MANUFACTURER   "ddB0515"
+#define BLEDIS_MODEL          "nRF52840 BBoard"
+
+//--------------------------------------------------------------------+
+// USB
+//--------------------------------------------------------------------+
+#define USB_DESC_VID           0x1d50
+#define USB_DESC_UF2_PID       0x6158
+#define USB_DESC_CDC_ONLY_PID  0x6157
```
hathach, line 60: could you provide the link where the vid is allocated to your board

hathach (commented):
thank you for you PR. It looks good, there is only a couple of minor questions.

.github/workflows/githubci.yml
ddB0515: no problem will change it

src/boards/nrf52840_bboard/board.h
ddB0515: here is PR on openmoko
https://github.com/openmoko/openmoko-usb-oui/pull/29

hathach: thanks for the link, both USB_DESC_UF2_PID and USB_DESC_CDC_ONLY_PID should be `0x6157`. Both are in bootloader mode, one with msc (uf2) and one without it (cdc only) for fast enumeration. MSC could take 2 more seconds to show up. Though it is safe to share PID between 2 modes.

ddB0515: sure thing, I didn't know that but will change it to follow your recommendation and use `0x6157`
Thanks

hathach (approved):
look great, thank you

PR #266, opened 2022-05-17
Fix ci

### LINKED ISSUE #265: Workflow-build fails with error code 2 while installing toolchains
### Operating System

Linux

### INFO_UF2.TXT

None, issue happens when compiling the bootloader.

### What happened ?

When compiling a bootloader with github actions the action fails with an error during toolchain installation. I will link my failed workflow run, as that will give more information than I could provide in text form:
https://github.com/ebastler/Adafruit_nRF52_Bootloader/runs/6452117714

This is for my modified workflow, where I removed all boards but a custom one - exact same result occurs with the default workflow config and no changes.

### How to reproduce ?

1. Fork repository
2. Run "build" workflow
3. It will fail at "Install Toolchains"

### Debug Log

_No response_

### Screenshots

_No response_

hathach:
fix #265 
- update action script
- use toolchain cache instead of node xpm
- also update gcc from 9.3.1 to 10.2.1

### DIFF
--- a/.github/workflows/githubci.yml
+++ b/.github/workflows/githubci.yml
@@ -74,7 +74,6 @@ jobs:
          path: linkermap
 
     - name: Set Toolchain URL
-      # 9.3.1-1.2.1 is xpack modified version which significantly increased compiled size to 6-7KB and cause flash overflow.
       run: echo >> $GITHUB_ENV TOOLCHAIN_URL=https://github.com/xpack-dev-tools/arm-none-eabi-gcc-xpack/releases/download/v10.2.1-1.1/xpack-arm-none-eabi-gcc-10.2.1-1.1-linux-x64.tar.gz
 
     - name: Cache Toolchain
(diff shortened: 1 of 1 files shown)


carlosperate:
This is a bit of self promotion of a GitHub Action I mantain, so apoligies if it's unwanted, and please do feel free to ignore the message. Hopefully is more useful than spam.

I was going through old GitHub notifications and noticed this, so just to let you know that with https://github.com/carlosperate/arm-none-eabi-gcc-action, which does automatic caching, something like this could condense to just:

```yaml
- uses: carlosperate/arm-none-eabi-gcc-action@v1
  with:
    release: 10.2.1
```
Obviously the current CI confirg works well, so I'm not suggesting it should be changed, just something that could be useful in the future 👍

hathach:
@carlosperate wow, your action looks great, lots of Adafruit repo need gcc and this will help to tidy the ci yml. I will try to start migrate some as starter. Thank you for the suggestion.

carlosperate:
Glad to hear that! Let me know if you have any questions, I'd be happy to help.

hathach:
sure will :100:

PR #267, opened 2022-06-23
Adds support for the Challenger 840 BLE board

PontusO:
This PR adds support for the Challenger 840 BLE board.
It also adds the possibility to add a board specific initialization function (extern_board_init()) to the board specific files. This is done by having a weak declared function in boards.c that can be overridden by the board specific files.

### DIFF
--- a/.github/workflows/githubci.yml
+++ b/.github/workflows/githubci.yml
@@ -34,6 +34,7 @@ jobs:
           - 'bast_ble'
           - 'bluemicro_nrf52833'
           - 'bluemicro_nrf52840'
+          - 'challenger_840_ble'
           - 'ebyte_e104_bt5032a'
           - 'ebyte_e73_tbb'
           - 'electronut_labs_papyr'
@@ -60,6 +61,7 @@ jobs:
     steps:
     - name: Setup Python
       uses: actions/setup-python@v3
+
       
     - name: Checkout Code
       uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


hathach:
look great, you should also add the board to githubci.yml as well for it to be included in the release binaries

.github/workflows/githubci.yml
```diff
@@ -60,6 +61,7 @@ jobs:
     steps:
     - name: Setup Python
       uses: actions/setup-python@v3
+
```
hathach, line 64: can you remove the extra empty line here and other 2 empty lines below as well.

hathach (commented):
one more last-minute request change, I still prefer to have board_ as prefix for the weak function. Could you rename it from extern_board_init() to board_init_extra()

src/boards/boards.c
```diff
@@ -93,6 +95,8 @@ void board_init(void)
 #if ENABLE_DCDC_1 == 1
   NRF_POWER->DCDCEN = 1;
 #endif
+  // Make sure any custom inits are performed
+  extern_board_init();
```
hathach, line 99: need to change as well.

PontusO: Sorry, to early in the morning for this stuff =)

hathach: no problems at all, me as well without coffee

hathach (approved):
look good to me, thank you for the pr

PR #283, opened 2022-10-21
nRF52840: MCU to MCU updates via activation pin and QSPI

m-mcgowan:
# Overview

This PR implements changes in the bootloader to enable MCU to MCU updates, in a similar manner to the built-in ROM bootloaders provided on STM32L4 and ESP32 devices. The host MCU can programmatically enter bootloader mode by asserting a DFU pin on reset, which places the device in DFU bootloader mode, with the DFU protocol running over UART rather than USB serial.

Additionally, the bootloader protocol has been extended with commands to allow erasing and writing to QSPI flash. This can be used to flash a pre-prepared FAT filesystem image to QSPI flash, such as for use with CircuitPython, enabling factory programming of CircuitPython scripts along with the CircuitPython runtime.

All changes are opt-in only and are intended to be fully backwards compatible, meaning no changes to boards that don't have the new configuration flags defined.

# Configuration Changes

Boards that support DFU activation via a pin should define `DFU_PIN_ACTIVATION=1`.

The button pin, DFU activation pin and FRST pin configurations now specify a pull direction to allow active high or low depending upon how that GPIO is used. The defaults reflect was was previously being used (`BUTTON_PULL`).

New optional defines `PIN_DFU_ACTIVATE` and `PIN_DFU_ACTIVATE_PULL` allow a particular board to define the DFU activation pin and pull direction. When these defines are present, DFU pin activation is compile-time enabled. These defines are present only for the Adafruit nRF52480 express board, which uses BUTTON_2 - feather pin D2.  This was previously the FRST signal, but factory reset functionality has been removed prior to this PR, and so it was safe to reuse this.

The define `DFU_EXTERNAL_FLASH` enables the protocol extensions for programming the QSPI flash. It is defined as 0 by default, and defined as 1 only for the adafruit_nRF52480_express board.

When DFU activation pin is enabled, both USART and USB CDC serial are both compiled in, and selected at runtime. When the DFU pin is not asserted, the bootloader functions as it did before - USB CDC and MSD is enabled for use with adafruit-nrfjprog and UF2. UART is still used by default for boards with the MCU subvariant `nrf52`, as was the case prior to this PR.

## Other Implementation Details

* I added some critical sections around the hci memory pool and hci transport code, to prevent race conditions accessing shared state between the main program thread and code running on UART interrupts.

* When DFU is started via the GPIO, the device is not restarted when the application is flashed, in case the MCU wishes to flash more binaries, such as the QSPI filesystem. `bootloader_app_start()` previously both registered the app and launched the app. The function has been factored out into functions so the app can be registered without necessarily starting it.


bsatrom:
@hathach can you take a look at this PR, let us know if the changes are acceptable or if not, what we need to do to get it into a good spot for merging into main?

hathach:
sorry, I was a bit busy for now, will check this out whenever I could !!

bsatrom:
Hey @hathach have you had a chance to review this yet?

m-mcgowan:
@hathach Is there any way I can help you with reviewing this? I'm happy to walk you though the PR and talk through the changes, the motivation behind them etc.  It would be great to have this merged, so anything I can do to help just let me know. Thanks!

hathach:
sorry for late response, I am in the middle of reviewing this, however look like the additional code cause the flash to be overflowed. Let me think how to shrink the size of bootloader quickly first, then we can merge to this and continue with the PR.

```
LD feather_nrf52840_express_bootloader-0.7.0-16-g047531a-dirty.out
/home/hathach/.local/xPacks/@xpack-dev-tools/arm-none-eabi-gcc/11.2.1-1.2.2/.content/bin/../lib/gcc/arm-none-eabi/11.2.1/../../../../arm-none-eabi/bin/ld: _build/build-feather_nrf52840_express/feather_nrf52840_express_bootloader-0.7.0-16-g047531a-dirty.out section `.text' will not fit in region `FLASH'
/home/hathach/.local/xPacks/@xpack-dev-tools/arm-none-eabi-gcc/11.2.1-1.2.2/.content/bin/../lib/gcc/arm-none-eabi/11.2.1/../../../../arm-none-eabi/bin/ld: region FLASH overflowed with .data and user data
/home/hathach/.local/xPacks/@xpack-dev-tools/arm-none-eabi-gcc/11.2.1-1.2.2/.content/bin/../lib/gcc/arm-none-eabi/11.2.1/../../../../arm-none-eabi/bin/ld: section .bootloaderConfig LMA [00000000000fd800,00000000000fd857] overlaps section .text LMA [00000000000f4000,00000000000fd9e7]
/home/hathach/.local/xPacks/@xpack-dev-tools/arm-none-eabi-gcc/11.2.1-1.2.2/.content/bin/../lib/gcc/arm-none-eabi/11.2.1/../../../../arm-none-eabi/bin/ld: region `FLASH' overflowed by 728 bytes
collect2: error: ld returned 1 exit status
```

src/main.c
```diff
@@ -223,14 +258,27 @@ int main(void)
     // clear in case we kept DFU_DBL_RESET_APP there
     (*dbl_reset_mem) = 0;
 
-    // start application
-    bootloader_app_start();
+    // register the app
+    uint32_t app_addr = bootloader_app_register();
+    if (!dfu_activate) {
```
hathach, line 263: since we reset when `dfu_activate=1`, we can just skip the teardown and simply reset. Teardown is only needed when jumping to app. I will make changes accordingly.

hathach:
I need to push changes to your fork to update PR, would you mind give me the write permission ?

m-mcgowan:
> [hathach](/hathach)

Sure thing, this is done! Thank you for starting to review.

src/main.c
```diff
@@ -334,11 +391,20 @@ static void check_dfu_mode(void)
...
+      channel_teardown();
     }
   }
+
+  // Simply reset, allowing the host MCU to control the behavior on boot via the activation pin.
+  // The host MCU may keep the device in bootloader mode via the DFU activation pin when there are more
+  // binaries to flash (e.g. QSPI flash data)
+  if (PIN_DFU_ACTIVATE_PRESENT && dfu_activate)
+  {
+    NVIC_SystemReset();
```
hathach, line 406: reset immediately here with activation would reduce the modification of existing code. I don't have the set up to test, please double check to see if this commit stills work on your side. Then we can move on with the reviewing (it may take a couple more commits I think).

m-mcgowan: I understand the tear down isn't strictly necessary, but it does no harm either, right?!. :-) I would vote for leaving the code as is, since it's already well tested. I only changed the code that was strictly necessary to keep the changes to the bare minimum.

If you feel very strongly for the change then of course we can test it.

hathach: please do the testing again, reset here will allow us to keep the rest of main.c unchanged, we could also revert changes to the bootloader.c file as well. I will also make a couple more commits anyway mostly to moving/refactoring changes.

m-mcgowan: I will test this today and will post back with results. Thanks!

m-mcgowan: Unfortunately it didn't work. I can try to find out why later, but from the perspective of the programmer, everything was ok. All the packets were ACKed and it successfully programmed the CPy interpreter and an image for the flash filesystem. But on reset, the nRF stays in the bootloader. 

I'll look deeper into it and check the memory contents and look for other clues as to why it's not working.

hathach: hmm, that is weird, probably one of the teardown function is indeed required. I will also try to do some check as well

m-mcgowan: I'll review the changes to day to see if I have any suggestions.

m-mcgowan: What I'm seeing is that the app isn't "registered" with the bootloader. This is because of the Reset that you added when `dfu_activate` is `true`. This results in `bootloader_app_start()` never being called.  

That was the reason I separated registering the app from starting it.  The registration is still necessary so that the app is launched when the host MCU restarts the device without the DFU activate pin asserted.

m-mcgowan: I'm still seeing intermittent errors. It is like the application doesn't "stick".

m-mcgowan: I don't have the root cause, but we are able to make this work in cases when the application is the last binary flashed (so qspi filesystem then the app). When the application is flashed first, then the QSPI filesystem, the application isn't saved. I believe the error is on our end (most likely related to timing), because I see the same result when I revert the recent changes to the PR back to my original PR, which I know was working, so something else must have changed.  So I would say this PR is good to merge as is.

GlibSkunk:
I would also appreciate the DFU over UART pins for the Feather nRF52840 Express, so thanks for the work guys! One request though if possible... could a new "Magic Number" also be added to the GPREGRET which points directly to DFU over UART?

In my scenario I communicate with the feather through UART exclusively. In my application code, I would then put in a uart command to set GPREGRET to the "DFU_MAGIC_SERIAL_UART_ONLY_RESET" and trigger a reset. This would avoid requiring an additional pull on the DFU pin.

First time Github poster, so I hope this request isn't completely out of line.

PR #284, opened 2022-11-10
Add new board PillBug

kylemccreery:


hathach (commented):
thank you for PR, would you mind providing link or explain how you obtain the vid/pid

kylemccreery:
@hathach PID is purchased through MCS Electronics.

kylemccreery:
pid


This is the only way of verifying it that they can provide since they do not do publicly facing databases. I have edited out the address used with the purchase.

hathach (approved):
thank you for the response, it look great now

PR #289, opened 2023-01-25
Add support for E104-BT5040U

amenekowo:
Hello there! I can't find a bootloader for this dongle so I supported it myself.
Original board.h is from mdk-52840 dongle. OTA name and else is referred to ebyte_e104_bt5032a module. And I left USB VID and PID as original, don't know if it's ok.
Tested on my dongle, flash using openocd with cmsis-dap.


src/boards/ebyte_e104_bt5040u/board.h
```diff
@@ -0,0 +1,73 @@
...
+
+//--------------------------------------------------------------------+
+// USB
+//--------------------------------------------------------------------+
+
+// Shared VID/PID with Feather nRF52840, will be disabled for building in the future
+// Dunno it's appropriate, but left here for a while.
+#define USB_DESC_VID           0x239A
+#define USB_DESC_UF2_PID       0x0029
+#define USB_DESC_CDC_ONLY_PID  0x0029
```
hathach, line 66: shared VID/PID is not allowed, please ask for an VID/PID allocated specially for your board.

src/boards/ebyte_e104_bt5040u/board.h
```diff
@@ -0,0 +1,73 @@
...
+
+#define UF2_PRODUCT_NAME    "Ebyte E104-BT5040U"
+#define UF2_VOLUME_LABEL   "EBT5040U"
+#define UF2_BOARD_ID       "10290-v1.0" // Silk on the PCB
```
hathach, line 70: BOARD_ID must follow uf2 specs format, checkout uf2 specs or other board for the format.

hathach (changes requested):
thank you for the PR, though please update following review feedback.

src/boards/ebyte_e104_bt5040u/board.h
hathach: unfortunately, an unique VID/PID for each board is required.

PR #292, opened 2023-02-02
Support for no button

matsujirushi:
[Seeed Studio XIAO nRF52840](https://www.seeedstudio.com/Seeed-XIAO-BLE-nRF52840-p-5201.html) and [Sense](https://www.seeedstudio.com/Seeed-XIAO-BLE-Sense-nRF52840-p-5253.html) do not have buttons.
(Double-click Reset to transition to DFU.)

### DIFF
--- a/src/boards/boards.c
+++ b/src/boards/boards.c
@@ -44,6 +44,9 @@
 #endif
 
 //------------- IMPLEMENTATION -------------//
+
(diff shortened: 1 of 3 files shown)


JohnAZoidberg:
It has one button, the reset button. And like you said, double-pressing that gets the module into DFU

PR #293, opened 2023-02-12
Set the default max MTU size within the DFU transport to 23

### LINKED ISSUE #287: DFU OTA with NRF Connect stuck at 100%
### Operating System

Android 12

### INFO_UF2.TXT

-

### What happened ?

I do the DFU Update with NRF Connect App.
The Distribution Package (.zip) is generated by Arduino (NRF52 Adafruit).

9 of 100 times the Update fails at 100% and the NRF is bricked. -> i have to flash the Bootloader again.
What should i do?

### How to reproduce ?

Do the DFU Update with NRF Connect

### Debug Log

_No response_

### Screenshots

_No response_

Montvydas:
Changed the default max MTU size within the DFU transport to 23, which is the required size for the legacy bootloader to perform a successful OTA with the latest Nordic-DFU-Library. This resolves #287.

### DIFF
--- a/lib/sdk11/components/libraries/bootloader_dfu/dfu_transport_ble.c
+++ b/lib/sdk11/components/libraries/bootloader_dfu/dfu_transport_ble.c
@@ -37,7 +37,7 @@
 
 
 #define BLEGAP_EVENT_LENGTH             6
-#define BLEGATT_ATT_MTU_MAX             247
+#define BLEGATT_ATT_MTU_MAX             23
 enum { BLE_CONN_CFG_HIGH_BANDWIDTH = 1 };
 
 #define DFU_REV_MAJOR                        0x00                                                    /** DFU Major revision number to be exposed. */


hathach (commented):
I see, it seems to be the lack of memory issue with higher MTU. Though 23 is a bit slow, can you try with a few higher MTU such as 50, 70, or 100 to see if we could have an optimum MTU.

Montvydas:
So I think it is a bit different than that. As I mentioned on the issue, one can use an older version of nRF Connect App e.g. v4.24.3 and then the MTU can be left as 247 (or to be honest, any number). You see, I tried experimenting with different MTU sizes and nRF Connect versions and noticed this:
- With nRF Connect v4.24.3 I can set MTU to 15, 23, 50, 247 and it will work, although the average speed will always be 4 kB/s. Meaning, the MTU value within the Bootloader isn't used at all, it is set by the nRF Connect App.
- With nRF Connect v4.26.0 (latest) the upload speed will be different depending on the MTU value. With MTU=15 it will not even start uploading, MTU=23 the avg speed=2.8 kB/s, with MTU=50 the avg. speed=5.7 kB/s, with MTU=247 the avg. speed=~70 kB/s (Using this the update rate is choppy, so at first it updates 60% straight away, then another 20%, then 10%..). However only the MTU=23 will complete the firmware upload successfully, while the others will hang up after reaching 100% (or will not even start at all) and will never be able to boot into the application mode. 

I searched this up within Nordic-DFU-Library and I found some proof that the current Nordic DFU Library allows using higher MTU values than 23, but ['MTU > 23 is only supported in the official Secure DFU from SDK 15 onwards.'](https://github.com/NordicSemiconductor/Android-DFU-Library/pull/260#issuecomment-665550946), more on the matter [here](https://github.com/NordicSemiconductor/Android-DFU-Library/issues/134#issue-371026935) and [here](https://github.com/NordicSemiconductor/Android-DFU-Library/issues/369#issuecomment-1325570491). Knowing that the Adafruit nRF Bootloader is an older unsecure version we can safely assume that for this to work we have to set the MTU to 23. The only thing I am not fully sure is why when I set the MTU to 23 and use nRF Connect v4.26.0, the avg. firmware upload speed is 2.8 kB/s, which is lower than 4 kB/s, which can be achieved with nRF Connect v4.24.3. The only explanation is that something else changed within Nordic DFU Library, that affects this, which has nothing to do with the MTU size.

Note this does not affect the MTU speed when used within the application mode, I tested this using the throughput.ino example by requesting to change to different MTU values there:
```
MTU=23
Sending 20000 bytes ...
Sent 20000 bytes in 1.15 seconds.
Speed : 17.33 KB/s.
---
MTU=247
Sending 244000 bytes ...
Sent 244000 bytes in 3.34 seconds.
Speed : 73.05 KB/s.
```
 
Regardless, now we know that this fix will not affect the older version of the Nordic-DFU-Library (e.g. the one used with nRF Connect App v4.24.3), since it would enforces the MTU to be 23 regardless (meaning it is backwards compatible). Then, it will ensure that we can use Adafruit nRF52 Bootloader with the latest version of the nRF Connect App, nRF Toolbox and the Nordic-DFU-Library, meaning it will work with Android 12 and will also support Flutter!

hathach (approved):
thank you very much for your sufficient testing. Indeed this cross-version compatible is a bit headache, which explain why I haven't experienced the issue under my mobile setup and others has quite a bit of issue with this. Changing back to normal MTU=23 even though is a bit slow but seems to be the safest thing to do for now. When having more time to test with nrf mobile API, we will re-increase it if needded.

Montvydas:
> thank you very much for your sufficient testing. Indeed this cross-version compatible is a bit headache, which explain why I haven't experienced the issue under my mobile setup and others has quite a bit of issue with this. Changing back to normal MTU=23 even though is a bit slow but seems to be the safest thing to do for now. When having more time to test with nrf mobile API, we will re-increase it if needded.

Just to clarify, I went through the comments and discussions on the Nordic DFU Library github. I found out that the previous version of Nordic DFU Library was forcing MTU=23 regardless of the value set within the bootloader. This was due to both the Nordic DFU Library not being able to support higher MTUs nor the non-secure bootloader. Then, in the latest versions of the Nordic DFU Library they enabled support for higher values of MTUs, which work with a configured secure bootloader. So keeping this in mind, if we want higher MTUs in the future, the bootloader will have to become secure, however I understand that would require a whole rewriting of the bootloader, is that right?

Montvydas:
@rcarteraz Could you please share a bit more information on the versions that you have tested and what did you try e.g. can you fill this in:
-  With nRF Connect v4.24.3, without this patch _works_ / _doesn't work_ on iPhone/Macbook.
- With nRF Connect v4.24.3, with this patch _works_ / _doesn't work_  on iPhone/Macbook.
- With nRF Connect v4.26.0 , without this patch _works_ / _doesn't work_  on iPhone/Macbook.
- With nRF Connect v4.26.0 , with this patch _works_ / _doesn't work_  on iPhone/Macbook.

I am not sure if you can install an older version of nRF Connect on iPhone, but maybe on Macbook it should be possible finding an older version. When you do this, I could then take a look and see if there are any obvious fixes.

What I am trying to understand here is whether this setting breaks something that worked or instead it fixes the issue for the Android system, but does not fully fix for Apple Ecosystem.

Montvydas:
@rcarteraz So the DFU app uses the same library, so ultimately it's the same thing. I am simply not sure which version of DFU app did the updates the "old" way... I am sure it's possible to track this down from the github release list on nRF repo.

PR #294, opened 2023-03-01
Shrink cf2 to 1 kb

hathach:


hathach:
not needed for now, and probaly cause issue witb existing app

PR #319, opened 2024-01-05
Add slash screen

hathach:
## Description of Change

Add slash screen support when in bootloader mode for board with built-in TFT. Currently only support ST77xx controller type. Supported boards are:
- Clue
- feather sense TFT (tested with feather express + 1.14 inch LCD). Actual hardware with different rotation (MADCTL may need to be updated)

@ladyada

IMG20240105164501 (1)

IMG20240105190800


ladyada:
hi when you get the hardware could you test this PR? i definitely have the display working in arduino but not in bootload mode for some reason :/

hathach:
For sure, I will do, just filled out the custom form. It will probably be delivered on Monday or Tuesday

PR #328, opened 2024-05-09
add neopixl power pin for Feather840 revE

hathach:
## Description of Change

- add P1.14 as NEOPIXEL_POWER_PIN for feather nrf52840 revE
- bump up gcc to 12.3.Rel1
- remove linkermap from ci

@ladyada


ladyada:
thanks! verified this fixes it :)

PR #344, opened 2024-11-16
fix reboot loop - set REGOUT0 only when value is erased to default

fanoush:
This fixes reboot loop when REGOUT0 is explicitly set to different value (like 3.0V).

The main rationale for setting REGOUT0 UICR value at boot time is to correct the 1.8V default which is there when UICR is mass erased. However when user sets different value explicitly, current code causes never-ending reboot loop with board default of 3.3V because you can't set flash bits back from 0 to 1.

This code fragment is typical to change the value only if it is the default (all FFs), see other places
https://github.com/search?q=UICR_REGOUT0_VOUT_DEFAULT&type=code


fanoush:
actually when thinking about it, it would still cause loop if someone #defines UICR_REGOUT0_VALUE to be  UICR_REGOUT0_VOUT_DEFAULT. This makes little sense and other code I see don't handle such misconfiguration too, but to be 100% correct we could change the `#if` above it to be
`#if defined(UICR_REGOUT0_VALUE) && (UICR_REGOUT0_VALUE != UICR_REGOUT0_VOUT_DEFAULT)`

hathach (approved):
great fix, thank you

PR #354, opened 2025-05-09
fix: CMake build producing too large binaries

andreaskurz:
## Checklist

- [x] Please provide specific title of the PR describing the change

-----------

## Description of Change

This change will fix build errors with CMake as size optimization is missing and therefore it doesn't compile successfully. Also the documentation is extended to give a hint how to use CMake in this repository at all.


andreaskurz:
> thank you, I change to use default CMAKE_BUILD_TYPE=MinSizeRel instead of forcing -Os for all build type. Also update doc for make as default generator (which is what most people would do).

Ah, thanks :)
Just learned about that option last weak in a customer project and forgot to update here.

PR #360, opened 2025-06-12
feat: add Seeed XIAO nRF52840 BLE/Sense boards with SoftDevice S340

JuergenLeber:
## Checklist

- [X] Please provide specific title of the PR describing the change
- [X] If you are adding an new boards, please make sure
  - [X] `UF2_BOARD_ID` in your board.h follow correct format from [uf2 specs](https://github.com/microsoft/uf2#files-exposed-by-bootloaders)

-----------

## Description of Change

This PR needs #359 as a prerequisite and extends the two existing Seeed XIAO nRF52840 BLE and Sense boards with SoftDevice S340 functionality. The automatic build workflow for these two new boards will fail until #359 has been pulled as the closed source files are not included and S340 builds are then excluded. The build only works locally after following the instructions in the README.md to add the necessary closed source files. It doesn't break anything else.


hathach (commented):
you may want to include the ci _s340 filter here as part of the PR as well. Though I am still thinking whether we should include these board with s340 since it is not buildable (therefore not maintainable in the future). Since s340 is only local buildable, and need local changes, maybe it makes sense to let user change the board.mk from s140 --> s340 themself.

JuergenLeber:
You're right @hathach - every board with S340 would just be a copy of a S140 one with some name changes. What do you think about a small script that would do the job locally? (copy board, make changes in naming, etc)

hathach:
I think the only changes is needed is SD_NAME and SD_VERSION, we can change = in board.mk/cmake with ?= (ifndef) to allow passing it with make/cmake command line and that should be good enough ?

JuergenLeber:
@hathach I'll take care of it and will change this PR as discussed!

PR #374, opened 2026-01-14
Improve bootloader version parsing from GIT_VERSION

piotrmalek:
## Description of Change

This PR improves the parsing of GIT_VERSION in the Makefile when generating
the bootloader version value.

The previous logic assumed a strict MAJOR.MINOR.PATCH format and could produce
incorrect results for common Git tag formats such as:

- v1.2.3
- 1.2.3-<commit>

The new implementation extracts semantic version components more robustly
and safely defaults to 0.0.0 if the version string does not match the expected
format.

This change does not affect existing version formats and is backward compatible.


hathach (approved):
thank you, I have update cmake to also match makefile

PR #377, opened 2026-01-26
Try to gracefully handle FLASH operations queue becoming full, instead of silently failing and allow OTA DFU process to be recoverable, in case of a communication error in the DUAL_BANK configuration

ejtagle:
This pull request addresses 5 issues that were found while testing:

1) When trying to queue a new FLASH erase or write operation, sometimes, depending on timing, the queue of operations was full, leading to silent discarding of the operation on 'release' builds, or even bootloader crashes or misbehaviour, if the discarded operation was to update the bootloader configuration.
With this patch, if that condition is detected (the write/erase queue of commands was completely full), wait until space becomes available - This won't fix the case when BLE comms sends firmware packets faster that the bootloader is able to flash them (so, it is still required to use PRN=8), but makes the bootloader behave properly if uploading at fast speeds.

2) The other thing that was fixed is an error introduced by github copilot (a suggested refactoring) that was incorrect: On the function dfu_bl_image_validate, a fix for switching from a dual_bank to a single_bank was introduced. Basically (https://devzone.nordicsemi.com/f/nordic-q-a/16590/dfu-change-bootloader-to-single-bank), the problem with updating the bootloader from a dual to single bank bootloader is that the new bootloader (single_bank) expects the firmware image to be at the start of BANK 0, while it's actually loaded to BANK1 by the old bootloader (double_bank). After the update the new single bank bootloader verifies that the image was correctly written to flash by comparing itself with the firmware image that it expects to be located in BANK0. Since the actual firmware image is in BANK 1 this check will fail and the new bootloader updates itself with whatever data located in BANK0 using the MBR, and that happens in a loop, preventing the user application to be launched, and the only possible recovery of that situation is by using a JTAG programmer.
The fix is to modify the dfu_bl_image_validate() function run verification on both addresses (BANK0 and BANK1), and if ANY of both banks contains the bootloader, then avoid copying itself
Originally, the code read: 

return sd_mbr_command(&sd_mbr_cmd_1) && sd_mbr_command(&sd_mbr_cmd_2);

sd_mbr_command() will return 0 (=NRF_SUCCESS) if the comparision matches, and !=0 if it does not. So, any of both comparisons returning SUCCESS means the dfu_bl_image_validate() will return SUCCESS (=0) as we are dealing with a logical AND operation. Copilot wrongly refactored it as if both comparisons must be SUCCESSful for the function to return success.
This is also fixed by this patch

3) When a DUAL_BANK bootloader is being used, and an OTA DFU is being carried, and the firmware transfer is interrupted or fails (for any reason), the bootloader is left in a "wait for more firmware packets" state. The only way to recover from such state, and attempt a new firmware transfer, is to send the reset command (that is exactly what the Nordic DFU does in this situation), and it is expected the bootloader to reenter the bootloader mode after a system reset: This exactly what happens in a SINGLE_BANK bootloader, as the main application firmware was erased, and there is no user application to jump to. 
But on the DUAL_BANK bootloader case, there is a user application (the last previously working one), so, without this patch, the bootloader was reseting the system and the user application was launched, thus preventing the Nordic DFU app to complete the firmware transfer. This specific case is now handled and the bootloader will reset and restart in OTA DFU mode, allowing the OTA DFU to complete successfully

4) Allow the bootloader to default to OTA DFU instead of Serial/UF2 DFU (for field deployed devices where no user accessible USB ports are available) so devices are recoverable

5) Clear pending interrupts and exceptions before launching the application. Sometimes (it is rare but it happens) with Softdevice 6.1.1, the bootloader jumps to the application and the device resets. This seems to be caused by a pending interrupt/exception ... The SDK12+ bootloader cleans them before jumping, make this bootloader do the same

6) Add ability to use an external PA/LNA to increase BLE range (disabled by default, configurable from the board.h specific to the board targeted)

### DIFF
--- a/lib/sdk11/components/libraries/bootloader_dfu/dfu_single_bank.c
+++ b/lib/sdk11/components/libraries/bootloader_dfu/dfu_single_bank.c
@@ -789,8 +789,9 @@ uint32_t dfu_bl_image_validate(void)
         sd_mbr_cmd_1.params.compare.len  = bootloader_settings.bl_image_size / sizeof(uint32_t);
 
         uint32_t err_code = sd_mbr_command(&sd_mbr_cmd_1);
-        if (err_code == NRF_SUCCESS)
-            return NRF_SUCCESS;
+        if (err_code == NRF_SUCCESS) {
+          return NRF_SUCCESS;
+        }
 
         uint32_t bl_image_start_2 = (bootloader_settings.sd_image_size == 0) ?
                                   DFU_BANK_1_REGION_START :
--- a/lib/sdk11/components/libraries/bootloader_dfu/bootloader.c
+++ b/lib/sdk11/components/libraries/bootloader_dfu/bootloader.c
@@ -140,7 +140,7 @@ static void wait_for_events(void)
         (m_update_status == BOOTLOADER_SYS_RESET) ||
         (m_update_status == BOOTLOADER_RESET_TO_SELF))
     {
-      // When update has completed or a timeout/reset occured we will return.
+      // When update has completed or a timeout/reset occurred we will return.
       return;
     }
   }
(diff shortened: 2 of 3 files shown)


lib/sdk11/components/libraries/bootloader_dfu/dfu_single_bank.c
```diff
@@ -753,9 +789,8 @@ uint32_t dfu_bl_image_validate(void)
         sd_mbr_cmd_1.params.compare.len  = bootloader_settings.bl_image_size / sizeof(uint32_t);
 
         uint32_t err_code = sd_mbr_command(&sd_mbr_cmd_1);
-        if (err_code != NRF_SUCCESS) {
-          return err_code;
-        }
+        if (err_code == NRF_SUCCESS)
```
hathach: right, spot on. it is my bad.

Issue #373, opened 2025-12-20
Compile error - array subscript


greifer500:
### Operating System

Linux

### INFO_UF2.TXT

Not available

### What happened ?

Same thing in [Issue 339](https://github.com/adafruit/Adafruit_nRF52_Bootloader/issues/339)

### How to reproduce ?

1. Install [arm toolchain](https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads)/arm-gnu-toolchain-15.2.rel1-x86_64-arm-none-eabi.tar.xz
2. paste content into /usr/local/bin
3. add "export PATH="$PATH:/usr/local/bin/arm-gnu-toolchain-15.2.rel1-x86_64-arm-none-eabi/bin""
4. Clone repo
5. cd into repo
6. make BOARD=pca10056

`ifneq (,$(filter 12.% 13.% 14.%,$(shell $(CC) -dumpversion 2>/dev/null))) CFLAGS += --param=min-pagesize=0 endif`

fixed by adding 15.%
`ifneq (,$(filter 12.% 13.% 14.% 15.%,$(shell $(CC) -dumpversion 2>/dev/null))) CFLAGS += --param=min-pagesize=0 endif`

### Debug Log

`CC bootloader_settings.c
lib/sdk11/components/libraries/bootloader_dfu/bootloader_settings.c: In function 'bootloader_mbr_addrs_populate':
lib/sdk11/components/libraries/bootloader_dfu/bootloader_settings.c:45:7: error: array subscript 0 is outside array bounds of 'const uint32_t[0]' {aka 'const long unsigned int[]'} [-Werror=array-bounds=]
   45 |   if (*(const uint32_t *)MBR_BOOTLOADER_ADDR == 0xFFFFFFFF)
compilation terminated due to -Wfatal-errors.
cc1: all warnings being treated as errors
make: *** [Makefile:424: _build/build-pca10056/bootloader_settings.o] Error 1`

hathach:
fixed by #381

PR #381, opened 2026-01-28
fix: Fix build errors on GCC15

ReFil:
## Checklist

- [x] Please provide specific title of the PR describing the change

-----------

## Description of Change

Much like with GCC 12, 13 and 14 GCC 15 will throw array initialisation errors during compilation when performuing register accesses in bootloader_settings.c. GCC 15 also throws an error around the handling of string like non-string elements in ghostfat.c. This PR adds gcc 15 to the workaround versions in the makefile to fix the first error and adds a new nonstring define which resolves to blank if the attribute isn't available for the second

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -392,12 +392,6 @@ endif
 
 CFLAGS += -DDFU_APP_DATA_RESERVED=$(DFU_APP_DATA_RESERVED)
 
-# https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105523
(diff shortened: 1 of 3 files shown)


ReFil:
fixes #373

hathach (approved):
thank you for your PR. I made minor adjustment
- added pragma GCC diagnostic ignored, that we fixed this once and for all. 
- Also corect usage of __has_attribute(nonstring) and rename it to ATTR_NOSTRING

Issue #21, opened 2018-08-15
use Access Control List (ACL) to protect bootloader


hathach:
Application can flash/erase bootloader code by accident. Use ACL to prevent such tragedy.

tannewt:
When we add this lets also add the ability to skip it when one of the retained registers is a specific value. That way we can self-update the bootloader.

hathach:
implemented by #384

PR #384, opened 2026-02-02
Use ACL to protect MBR and Bootloader against accidental FLASH overwrites by the user application

ejtagle:
This pull request makes the bootloader protect the FLASH areas where the MBR and the Bootloader is stored against overwriting by the user application. This should address the following issue: https://github.com/adafruit/Adafruit_nRF52_Bootloader/issues/21

I am not absolutely sure if we want or not this feature... Basically, this makes impossible to write a user application that updates the bootloader, so, for some specific user scenarios, we actually don't want this feature.

But, for most of the others, i think this feature is desirable, as, let's be honest here, most users are just amateur coders and a simple error in their flash writing routine will make the device unrecoverable without using a JTAG programmer...

The other thing is that this protection is enabled by default, changing the "de facto" behaviour of the previous bootloader versions... so, I am not sure... I think enabling the protection is simply better for most of the cases

I have tested this very quickly, so, also, i would appreciate if anyone can test it a little bit more.

Why not protecting the softdevice ? ... I think protecting the bootloader would be enough, but, protecting the softdevice is also possible (I don't know if the softdevice itself writes to FLASH ... that is something that should be tested, but it is suspicious SDK12+ secure bootloaders do not protect the softdevice, so, I just followed what they did there

I am open to suggestions and changes, or even dropping the pull request. For my specific use case, enabling protection seems the most sensible thing to do, and also, on the newer SDK12+ secure bootloaders, this protection is always done ..

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -397,12 +397,6 @@ endif
 
 CFLAGS += -DDFU_APP_DATA_RESERVED=$(DFU_APP_DATA_RESERVED)
 
-# https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105523
-# Fixes for gcc version 12, 13 and 14.
-ifneq (,$(filter 12.% 13.% 14.%,$(shell $(CC) -dumpversion 2>$(NULL_DEVICE))))
-	CFLAGS += --param=min-pagesize=0
-endif
-
 #------------------------------------------------------------------------------
 # Linker Flags
 #------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


hathach:
> The SDK12+ also enables write protection of the MBR/Bootloader from the Bootloader itself (if you inspect the main() function of this SDK12+ based secure bootloader (https://github.com/kaidyth/nrf52_bootloader/blob/master/main.c , lines 216 to 221), the first 2 lines executed do that), but this means the bootloader itself cannot be updated from itself, and/or the MBR cannot be initialized the first time, so those lines do not make any sense to me at all (yeah, it would be great to protect the bootloader/softdevice from unintended corruption from the bootloader code itself, but once the write protection is enabled, according to the technical docs, it is impossible to disable) According to this post [probe-rs/probe-rs#364](https://github.com/probe-rs/probe-rs/issues/364) , the softdevice itself contains checks that will forbid the user app to modify/erase parts of it, so I suppose all corner cases should be covered

since we only enable ACL when launching application. We can keep the self-update feature right ? please make sure you test it on actual hardware.

PR #390, opened 2026-03-08
Exit bootloader and jump to app when USB is disconnected

jesuslg123:
## Summary

This PR updates bootloader startup behavior so it automatically jumps to the application in specific USB-related cases, provided the application is valid.

## New behavior

The bootloader now jumps to the app when:

- USB is not enumerated within 3 seconds during boot.
  - If no USB host enumerates the device in that window, the bootloader times out and starts the app.
- USB was enumerated during startup, but is later physically disconnected.
  - In that case, the bootloader exits startup DFU and starts the app.

In both cases, the jump only happens if a valid application is present.

## Behavior that does not change

This auto-jump logic does not apply to DFU flows that are intended to keep the device in DFU mode, including:

- Double-reset DFU
- DFU button entry, or DFU+OTA button entry
- OTA/BLE DFU
- Any path that calls `bootloader_dfu_start(..., 0, false)`

## Rationale

This makes USB-startup DFU less sticky for normal boot scenarios: if a host never enumerates the device, or if the USB cable is disconnected after enumeration, the device can continue into the main application automatically instead of remaining in the bootloader unnecessarily.


hathach (approved):
perfect, thank you. I tidy it up a bit by setting was enumerated using tud_mount_cb(). The rest look good.

